PR #4, opened 2020-10-09
SQL Server ODBC: set length for varchar parameters

pavelmash:
see https://github.com/synopse/mORMot/pull/359


synopse (approved):
Thanks a lot for pushing the modification also to the mORMot2 repository!

PR #5, opened 2020-11-22
Add missing unit

JavierusTk:
Compilation under Delphi 10.3.3 stops with:
[dcc32 Error] mormot.db.rad.firedac.pas(322): E2003 Undeclared identifier: 'TDynArray'
Adding mormot.core.data, to the uses clause solves it


synopse:
Thanks for the feedback!

PR #6, opened 2020-11-23
HTTP_DEFAULT_ACCESS_RIGHTS implementation change

JavierusTk:
If this unit is compiled as a runtime pacakage, and the  package does not contain mormot.orm.core.pas (where  SUPERVISOR_ACCESS_RIGHTS is located), then compiling will fail with:
[dcc32 Error] E2201 Need imported data reference ($G) to access 'SUPERVISOR_ACCESS_RIGHTS' from unit 'mormot.rest.http.server'

### DIFF
--- a/src/rest/mormot.rest.http.server.pas
+++ b/src/rest/mormot.rest.http.server.pas
@@ -122,8 +122,11 @@ type
 
 const
   /// the default access rights used by the HTTP server if none is specified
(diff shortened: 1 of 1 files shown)


synopse:
I guess this won't be enough for other part of the framework, so including mormot.orm.core.pas to the package is the right way to go. Note that we don't support packages yet with mORMot 2.

JavierusTk:
I've already sliced it in packages, and this one is the only needed change to make it all compile

PR #7, opened 2020-11-23
two errors related to SendEmail

JavierusTk:
[dcc32 Error] mormot.net.client.pas(33): E2029 Declaration expected but identifier 'mormot' found
Change ";" for "," in USEWININET uses

[dcc32 Error] mormot.net.client.pas(2203): E2250 There is no overloaded version of 'GetNextItem' that can be called with these arguments
Change P: PAnsiChar;   forh   P: PUTF8Char;


synopse:
Had already been fixed.

Please wait a little when something is wrong from a latest commit.

PR #8, opened 2021-01-24
Changes for Android.

LongDirtyAnimAlf:
This pull includes changes for Android. With these changes, I get all greens on armhf , armel and arm64.
Most work went into armel: no fpregs are used. Modern Android for arm (NDK>R14) uses armel !!

Please review the changes.

As you can see, I used a console redirect to get the output on the screen of an Android phone.
This might be included in a more elegant manner in the final sources.


LongDirtyAnimAlf:
My advice would be not to accept this pull, but to merge all the changes into mORMot2 (and perhaps also some into mORMot1).

synopse (commented):
> Modern Android for arm (NDK>R14) uses armel !!
So the newer NDK uses an older architecture without hard floats? Weird.

LongDirtyAnimAlf:
So the newer NDK uses an older architecture without hard floats? Weird.
https://android.googlesource.com/platform/ndk/+/353e653824b79c43b948429870d0abeedebde386/docs/HardFloatAbi.md

imv* is the right prefix for mORMot 2
Didn't know for sure. In that case, all asm-calls have to use the imv*

Why? ARM has plenty of registers IIRC.
I do not know how the compiler assigns and uses the registers, but this change solved some errors on arm.

I don't get it. A constant is always aligned... But BuildDatetime may not.
As this only happens with FPC 3.2, this might also be a compiler error. But adding unaligned did not solve the sigsegv. Using a constant did !

Why use an 'if ... in' here?
Use whatever you find suitable. I just changed the code to fit the defines more easily. And perhaps prevent a jump-table due to the single-case statement.

What it is for?
Some time ago, we changed the armfakestub from mixed asm towards pure asm to prevent errors caused by compiler optimizer and a lack of a free register to transport the method index. The pure asm was more or less a copy of the mixed asm by looking at the generated asm code. The mixed asm is still there, together with the code to call this mixed asm. It is protected by {$ifdef ASMORIG} is the sources. The {.$define ASMORIG} was a left-over from my efforts to port towards armel.

Nothing to change in mormot.defines.inc ?
Nop. Most important changes are in mormot.core.interfaces.

I must say that the new mORMot2 is a joy to work with. Code is very easy to locate. And using FPC typinfo makes life also a lot easier.

synopse:
> https://android.googlesource.com/platform/ndk/+/353e653824b79c43b948429870d0abeedebde386/docs/HardFloatAbi.md

Now I get it: the ABI doesn't use the float registers, but the float calculation is still done in hardware.
Their arguments make perfect sense.

> I do not know how the compiler assigns and uses the registers, but this change solved some errors on arm.

Compiler bug then... the only thing which may be sensitive is the "goto" but we use it sometimes for better code generation, with no trouble. If it fixes the issue, I trust your patch! Perhaps a code review may help to identify if similar compiler issues could remain.

synopse:
A good read for me: https://wiki.freepascal.org/ARM_compiler_options#Interactions
So if ARMEL is needeed for newer Android, we should make a reference to the NDK note in the source code, and also to this wiki FPC page. I will do that once you give me green light for the merge!

LongDirtyAnimAlf:
Yes, it would be good to make a reference to the NDK note.
This reference is also included in fpcupdeluxe, for me to not forget.

About fpcupdeluxe.
It always builds an arm cross-compiler with -dFPC_ARMHF. However, the arm crosscompiler emits armel for Android and armhf for Linux. How smart.

armhf

I would welcome a merge now. And I think that finetuning might still be needed.

Next target: FPC trunk with attributes ... :-)

LongDirtyAnimAlf:
Hi Arnaud,

I have added the changes to get mORMot2 running with FPC trunk (and stable) on win, linux and arm.
All changes can be merged now. Work is finished.
I guess Delphi is not affected by this, but I did not check.

Thanks, Alfred.

synopse (approved):

PR #9, opened 2021-02-07
added script to build tests under Linux

pavelmash:
- added script to build tests under Linux
 - script to update static asset to latest
 
 Changes compared to Windows build scripts:
  - I use UPPERCASE for env variables what can be passed from outside word - this is a naming convention in Unix.
  - in  `get_latest_static.sh` I can't use a https://api.github.com/repos/synopse/mORMot2/releases/latest API call because of "pre-releaes" not works in this way, so I parse a JSON to get a latest pre-release
  - I use `build` instead of `compil` for naming a build script

Having such a scripts allows to setup a github action in future for CI 

Verifed on Ubuntu 20.04

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,8 @@
 # Uncomment this if you are not using external objects (zlib library for example).
 #*.obj
 #
+# static files (can be downloaded form mORMot 2 repositore releases page
+static/*/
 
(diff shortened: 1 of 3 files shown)


synopse (approved):
This is a great script!

PR #10, opened 2021-02-08
added share interface support for LibCurl (back-port form mORMot1 pr376)

pavelmash:
Back-port from https://github.com/synopse/mORMot/pull/376 with subsequent changes.

Also added misisng in mORMot2 call of curl.global_cleanup in TLibCurl destructor to ensure all shared resources are disposed correctly (sockets are closed, for example)


synopse (approved):
Great!

PR #13, opened 2021-03-01
Create lpk for lazarus with all units of mORMot

eblaudy:


synopse (approved):
I will probably rename it from mormot.pas into mormot2.pas since we already have a mormot.pas unit in mormot 1.18.
Thanks for the input!

PR #16, opened 2021-03-02
Sqlite api update

OkobaPatino:


synopse:
See https://synopse.info/forum/viewtopic.php?pid=34499#p34499 for additional info.

PR #18, opened 2021-03-03
Rearranged SQLite API, added some missing and cleanup

OkobaPatino:


synopse:
I am not sure I followed the pertinence of all changes.
Perhaps the units were changed since the time of your initial pull request.
Could you please take a look at the current state, and propose a minimal content?

For instance, I am not sure adding "sqlite_exec() return code" in the documentation of SQLITE_* constante is accurate. Those error codes are returned by much more than sqlite3.exec().

If you just want to add 3 new API functions, please just add them in your pull request.

OkobaPatino:
The patch is the follow-up of this post: https://synopse.info/forum/viewtopic.php?pid=34518#p34518
It contains rearranging APIs, fixing some typos, and some cleanups so it seems big.
I'm afraid I do not know what you mean by minimal content.
Also, I did not add "sqlite_exec() return code", I removed it as the reason you mentioned, the comment is for much older versions.

synopse:
As far as I can tell, there are some problems in the commit, e.g. about the order of functions which changed in the class pointers, but not as names.

But it is very valuable, so I will merge it and fix it if needed.

Thanks for sharing!

OkobaPatino:
Thank you. It is somewhat old and I do not remember the changes correctly, next time please let me know if there was any problem so I can fix it instead of giving you the trouble.

synopse:
It was not obvious that the SQLITE3_ENTRIES[] order should match the exact order of functions pointers.
So the dynamic loading was broken.

Should be fine now.

PR #19, opened 2021-03-03
Remove notes and hints messages at compilation of the lpk project

eblaudy:


synopse:
I guess the latest version remove the hints/warnings by number, which is a safer approach.

PR #22, opened 2021-03-13
2 more examples

martin-doyle:
- DDD
I am not sure if I understood your DDD approach correctly and would appreciate your feedback. 

- Docker
I find running the server in a Docker container very convenient. Tested for Linux x86_64 and ARM64.


synopse:
I merge the branch since it is highly educational, even if the code was not officially proven.

PR #23, opened 2021-03-23
Add   mormot.lib.static to uses

devaex:
Need for Before/AfertLibraryCall.


synopse:
Please don't use this unit - it is not yet workable.
An eventual commit will enable it.

PR #24, opened 2021-03-25
fix crypt units rename/move issue of lazarus pkg

uian2000:


synopse (approved):
Still missing the src/script folder but I will add it.

PR #27, opened 2021-04-20
[FIX] DocVariant DispInvoke:  wrong order of arguments when compiled …

htouvet:
…with fpc 3.2

(looks like order off arguments in CallDesc has been reverted since fpc 3.2)


synopse:
Has been implemented with a runtime test by https://github.com/synopse/mORMot2/commit/3fbd616683645e7eee14f894e4637cbc0787e21a

Thanks for the feedback!

PR #28, opened 2021-04-23
Fixed combined signed-unsigned warnings

sakura1977:
Combining of UInt64 and Int64 fixed


synopse:
on which compiler does it happen?

PR #30, opened 2021-05-02
Update mormot.core.rtti.pas

danielkuettner:
Hi AB,

in SetFloatProp there is a 

Value := 0;

after begin.
Please remove this line, then all works as expected.

Daniel


synopse:
Thanks for the feedback.

PR #39, opened 2021-07-26
Deleted duplicate "implementation" clause

JavierusTk:
I do have NexusDB, so when I compiled it, compiler complained


synopse:
Thanks for the feedback!

PR #40, opened 2021-07-27
Sqlite fixes

OkobaPatino:
Removed extra param, and changed unsigned int types to cardinal.


synopse (approved):
The sqlite3_result_pointer() issue is a nice catch. 8-)

Thanks for sharing!

PR #41, opened 2021-07-29
WGet: resume: if a .part file already exists but a a Size which is sa…

htouvet:
…me or above expected size, we should drop it.

### DIFF
--- a/src/net/mormot.net.client.pas
+++ b/src/net/mormot.net.client.pas
@@ -1652,6 +1652,7 @@ var
   res: integer;
   partstream: TStreamRedirect;
   resumed: boolean;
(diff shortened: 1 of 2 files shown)


src/net/mormot.net.client.pas
```diff
@@ -1652,6 +1652,7 @@ function THttpClientSocket.WGet(const url: RawUtf8; const destfile: TFileName;
   res: integer;
   partstream: TStreamRedirect;
   resumed: boolean;
+  ExpectedSize: LongInt;
```
synopse, line 1655: The problem is that the size is likely to be Int64...

synopse (approved):

PR #44, opened 2021-08-03
New instruction for adding library paths in Delphi

zedxxx:


zedxxx:
Using the new instruction you can simply copy-paste library paths to the IDE regardless of where the sources were copied to.

synopse (approved):
Yes, this readme is not finished yet.

At least it is better with your commit.
Thanks! :)

PR #46, opened 2021-08-22
BUGFIX - some fpc project adjustments

martin-doyle:
- Add crypt directory to the project path
- Remove PUREPASCAL flag


synopse:
Shouldn't in fact the mORMot package be used instead of manual paths for lazarus?

synopse (approved):

PR #47, opened 2021-08-24
TDynArray.Copy: Do not dereference PDynArray arguments

transmogrifix:
TDynArray.Copy: Do not dereference PDynArray arguments when passing them to DynArrayCopy, they might be nil and cause an access violation.
I encountered this issue when TSQLPropInfoRTTIDynArray.CopySameClassProp tried to copy a TRawUTF8DynArray where both src and dst were empty.


synopse (approved):
Nice catch.
Thanks for the feedback!

PR #48, opened 2021-09-03
init Value

danielkuettner:
otherwise in debug Value=0 but in release mode Value = 1 (Delphi 10.4, Win64 Release)


synopse:
This is as designed.
If the function returns false, the Value content is undefined.

danielkuettner:
Ok, I’ve dropped the pull request and understand this.

Daniel

PR #49, opened 2021-09-11
Use correct FlattenedProps in TOrmPropInfoRtti.CopyProp.

transmogrifix:
Destination FlattenedProps were not being used.


synopse:
I made a refactoring of your solution, for slightly better performance and maintainability.

PR #50, opened 2021-09-11
Use correct ORM props in TOrm.RttiJsonWrite.

transmogrifix:
Calling OrmProps directly returned props for TOrm, not the actual instance class OrmProps. 
I noticed it because the TOrm class doesn't have any fields, so I got an access violation when the method was called from `mormot.core.json._JS_RttiCustom` (`TOnRttiJsonWrite(nfo.fJsonWriter)(c.W, Data, c.Options)`).


synopse (approved):
This was a weird issue to catch.

Thanks for sharing!

PR #56, opened 2021-09-24
TOrmPropInfoRttiDynArray.SetValue: Try to base64magic decode T*ObjArrays as well.

transmogrifix:
The ORM saves my T*ObjArray property as a base64magic encoded blob. I noticed that the property was empty when loaded from the DB.

### DIFF
--- a/src/orm/mormot.orm.base.pas
+++ b/src/orm/mormot.orm.base.pas
@@ -6354,8 +6354,13 @@ begin
     da.Clear
   else
   try
-    if (fObjArray = nil) and Base64MagicCheckAndDecode(Value, tmp, ValueLen) then
-      da.LoadFrom(tmp.buf, PAnsiChar(tmp.buf) + tmp.len)
+    if Base64MagicCheckAndDecode(Value, tmp, ValueLen) then
+    begin
+      if fObjArray = nil then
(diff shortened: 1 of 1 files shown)


synopse (changes requested):
The ORM should not save the T*ObjArray as Base64 but as JSON.
So the fix is to be done at saving, not at loading.

transmogrifix:
I wasn't sure, thanks! I'll look into it.

transmogrifix:
Actually, the data in the DB is not Base64, it is JSON. I missed it because it is in a `bytea` column (PostgreSQL) and pgAdmin4 does not show content, just displays `[binary data]`. 
So it is being base64 encoded somewhere after it is fetched from the DB.

transmogrifix:
It is being Base64 encoded in `TSqlDBPostgresStatement.ColumnsToJson(WR: TJsonWriter)` because column type is ftBlob.
I am unsure how to proceed. OrmFieldType for this property is oftBlobDynArray which maps to ftBlob.

synopse:
It should be ftUtf8 for T*ObjArray.
Please try with my latest commit above.

transmogrifix:
That fixed it. The column is now text, not bytea so the issue no longer exists.
The refactored code looks much better. 
Thank you!

PR #57, opened 2021-09-26
Firebird ibx driver update/bugfix

TTomas:
Also added converted Sample 15-External DB performance with added ibx test.
Conversion is not PUREMORMOT2

### DIFF
--- a/src/db/mormot.db.sql.ibx.pas
+++ b/src/db/mormot.db.sql.ibx.pas
@@ -16,6 +16,10 @@ unit mormot.db.sql.ibx;
 
   For low level connection, uses the MWA Software Firebird Pascal API package,
   (fbintf) part of IBX for Lazarus - https://www.mwasoftware.co.uk/fb-pascal-api
+  This version work with trunk version of IBX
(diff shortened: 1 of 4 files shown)


src/db/mormot.db.sql.ibx.pas
```diff
@@ -189,17 +189,29 @@   TSqlDBIbxConnection = class(TSqlDBConnectionThreadSafe)
 
   /// implements a statement via a IBX/FB Pascal API database connection
   TSqlDBIbxStatement = class(TSqlDBStatementWithParamsAndColumns)
+  type
```
synopse, line 192: This code won't compile with Delphi 7 or 2007.

synopse (changes requested):

src/db/mormot.db.sql.ibx.pas
TTomas: I move sub-type, just to inform you MWA fbintf package minimum requirements:
FPC version 3.0.0
Delphi 2010+

synopse: Ah I didn't know that.
I am not sure the sub-type would compile with Delphi 2010.

PR #66, opened 2021-11-30
Small fixes for Windows.

LongDirtyAnimAlf:
Please review.
Some of these fixes might need an alternative fix.


LongDirtyAnimAlf:
Thanks for looking into the issues. And solving them.
Sidenote.
As this pull was based on issues caused by the mORMot2 running on FPC trunk, I guess that the inline issue with zip Size will still be present. But that is of no importance

PR #68, opened 2021-12-22
Fix AddInterceptors call parameter in ExecuteMethod

tihorygit:


synopse:
Thanks for the feedback!

synopse:
Just a problem when compiling with Delphi.

PR #71, opened 2022-01-08
Use mormot2 package

martin-doyle:


synopse:
Nice!
Thanks for the commit and feedback.

PR #74, opened 2022-02-12
Update mormot.core.os.pas

danielkuettner:
under Delphi Win32/64 there are av's, when doing this in TSQLHttpServer.HttpThreadStart:

fLock.Lock; --> here AV
try
...
finally
  fLock.Unlock;
end;

fLock is a private field of instance of TSQLHttpServer here.

### DIFF
--- a/src/core/mormot.core.os.pas
+++ b/src/core/mormot.core.os.pas
@@ -5871,6 +5871,8 @@ end;
 
 procedure TSynLocker.Lock;
 begin
+  if fSection.LockCount<0 then
+    Exit;
   case fRWUse of
     uSharedLock:
       begin


synopse:
I don't understand the problem and where this fLock is defined.

In TRestHttpServer.HttpThreadStart there is no fLock, but a fSafe: TRWLightLock

src/core/mormot.core.os.pas
```diff
@@ -5871,6 +5871,8 @@ procedure TSynLocker.RWUnLock(context: TRWLockContext);
 
 procedure TSynLocker.Lock;
 begin
+  if fSection.LockCount<0 then
```
synopse, line 5874: This is very specific to Windows, and won't compile with Linux/POSIX.
I don't understand what you want to achieve.

synopse (changes requested):

danielkuettner:
fLock is declared by myself, not at mORMot2 side.

I’ve used this SynLocker to protect my code in this method.

synopse:
If what you want to achieve is to have your own lock, and let it be used before it is initialized, it is not the good solution.
You should initialize the lock an overriden constructure, before calling inherited Create, i.e. before the Execute method is called.
Note that there is already a TRWLightLock within this class.
Please use the forum to discuss this.

danielkuettner:
I've just updated mORMot2 and the issue didn't exists before.
TRWLightLock was unknown for me (up to now).
I'm also not aware of constructor is called after HttpThreadStart is fired, so my fLock.Init in constructor was wrong.

PR #75, opened 2022-02-21
- Implement Firebird 4 IBatch interface for Batch Insert and Update

TTomas:
- Implement Firebird 4 IBatch interface for Batch Insert and Update
- Some fixes for current trunk and performance


synopse (approved):
Great job!

PR #76, opened 2022-02-28
Migration ex 31 chat-server/client

uian2000:
- change use units from mormot1 to mormot2
 - release success in m1 but error in m2
 - my env win10-x64 delphi xe/ lazarus 2.2(fpc3.2.2)


synopse (approved):
thanks for the conversion

PR #77, opened 2022-03-22
Draft: Try to fix SQL string corruption.

transmogrifix:
The SQL string becomes corrupted when there are several threads executing the same query with different parameters (not sure if that matters). Here is what the SQL string looks like:
image

I am not sure if this is where the problem is, but it does fix the issue. Probably several threads are using the same fTempBuffer^ at the same time.

I suspect the regression was introduced somewhere around commit 1d6bd19b34020648cd5e726513a547e9eb1c7299
Could you take a look?

Thank you.

### DIFF
--- a/src/orm/mormot.orm.sql.pas
+++ b/src/orm/mormot.orm.sql.pas
@@ -792,7 +792,7 @@ begin
       else
         FormatUtf8(limit.InsertFmt, [Stmt.Limit], limitSQL);
     end;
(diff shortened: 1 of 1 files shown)


synopse:
You are right.
The shared fTempBuffer should be protected by the StorageLock of the class.
I have made a code review, and https://github.com/synopse/mORMot2/commit/a6505de9 should fix the issue.
Sorry for the regression - I hope you did not lost too much table on it.

PR #78, opened 2022-03-23
fix regression after 96c8a5d3

danielkuettner:
https://synopse.info/forum/viewtopic.php?id=6180


synopse:
thanks for the feedback!

PR #79, opened 2022-03-25
Added several curl_mime_* methods.

transmogrifix:


synopse:
Thanks for the input!

synopse:
Thanks for the input!

synopse:
Your pull request needed some fixes for Android, i.e. for libcurl static linking.

transmogrifix:
Yes, not sure how I missed that. Probably because it was disabled due to the conditional.
Thank you!

PR #80, opened 2022-03-27
Fix output directory for the extraction of downloaded static files and make sure ONLYUSEHTTPSOCKET is honored in mormot.rest.http.server.

turkerali:
Fix output directory for the extraction of downloaded static files and make sure ONLYUSEHTTPSOCKET is honored in mormot.rest.http.server


synopse:
Note that get_latest_static.sh is deprecated and is better not used.

But I don't get why ONLYUSEHTTPSOCKET should use the old socket server: ONLYUSEHTTPSOCKET is fine with the async server, which uses sockets, and scales better.

PR #91, opened 2022-05-31
Fixed a locking issue in TSynCache 

transmogrifix:
A deadlock occures when ResetIfNeeded attempts to acquire a write lock which is already acquired by AddOrUpdate.


synopse:
Nice finding!

I think just using TRWLock instead should be enough, since its writes are reentrants.
Please see my above commit.

transmogrifix:
Even better, I didn't see that TRWLock is reentrant.
That is great, thank you!

PR #92, opened 2022-06-03
TLS support for THttpServer

achechulin:
Initial TLS support. 
Tested on Windows 10 (SSPI and OpenSSL 3.0.3), and in Ubuntu 22.04 (OpenSSL 3.0.3).
How to use:
```
fHttpServer := THttpServer.Create('443', nil, nil, '', 32, 30000, [hsoEnableTls]);
fHttpServer.WaitStarted; // raise exception e.g. on binding issue
fHttpServer.Sock.TLS.CertificateFile := 'cert.pem'; // cert.pfx for SSPI
fHttpServer.Sock.TLS.PrivateKeyFile := 'privkey.pem';
```


synopse:
I will try to implement it on the async server too.

achechulin:
Before, I looked at the NGINX source - async is much more complicated.

synopse:
my idea is to switch to synchronous/blocking socket mode for the TLS code

PR #97, opened 2022-06-19
Prevent sorting when resizing columns

rudiloos:


synopse (approved):
On Delphi? On FPC? On which version would it work?

PR #100, opened 2022-07-06
fix compile issue under Delphi 10.4

danielkuettner:
compile error was: "Konstantenausdruck erwartet"

### DIFF
--- a/src/lib/mormot.lib.openssl11.pas
+++ b/src/lib/mormot.lib.openssl11.pas
@@ -6300,14 +6300,14 @@ begin
 end;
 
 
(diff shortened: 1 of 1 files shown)


synopse (commented):
This compiles, but is not correct: @up does not point to the start of the constant but to the start of the location where the string pointer is stored, so pointer(up) should be use instead.
Anyway, I have fixed it in another way by https://github.com/synopse/mORMot2/commit/47477964

danielkuettner:
Thanks for explaining, and thanks for the great mORMot2!

I think nearly every day about you!

Daniel

PR #104, opened 2022-07-29
improve AddColumns speed - +500RPS on TFB /db endpoint

pavelmash:
Before this optimization AddColumns takes 2.17% (fpc_ansistr_concat_multy = 1.92%) CPU on TFB `/db`, after - 0.95. As a result +500 RPS


synopse:
I guess the output is not correct: there is a missing ':' character.

pavelmash:
Yes, ':' is missed - I add it

synopse:
Output is still incorrect.

I will make my own version.

pavelmash:
Closed in flavor of [2f060cf36]

PR #109, opened 2022-07-30
Improve HTTP header parser

pavelmash:
The idea is to move most useful headers up in IdemPPChar array.
This decrease CPU time spend in ParseHeader function from 17.5% to 15.9% (measured on /json TFB test using valgrind) and should help a little on real-life browser request.

/json RPS improved by 0.4%  ~500000 -> ~504000


synopse (approved):

synopse:
Good idea.

See also https://github.com/synopse/mORMot2/commit/3370b572 about another way of parsing headers.

PR #111, opened 2022-08-01
prefer mORMot StrLen over PQGetLength + prevent unnecessary PQGetIsNull calls

pavelmash:
`StrLen` is twice faster compared to `PQGetLength` 0.2% vs 0.4% on TFB `/db`


synopse:
Note: it is only for small buffers.
The source code of PQGetLength is pretty simple, and a single O(1) field lookup:
https://github.com/postgres/postgres/blob/master/src/interfaces/libpq/fe-exec.c#L3806

PR #112, opened 2022-08-01
Postgres: use StrLen only for small buffers, else PQ.GetLength

pavelmash:
As @synopse notes in #111


synopse:
Only exception is AddJsonEscape() because AddJsonEscape(P, 0) is faster than AddJsonEscape(P, somelen).

PR #114, opened 2022-08-02
Update mormot.db.nosql.mongodb.pas

danielkuettner:
same changes I've sent per Email (perhaps easier to check and merge)

### DIFF
--- a/src/db/mormot.db.nosql.mongodb.pas
+++ b/src/db/mormot.db.nosql.mongodb.pas
@@ -172,7 +172,7 @@ type
     // ! TMongoRequestQuery.Create('admin.$cmd','buildinfo',[],1)
     // will query   { buildinfo: 1 }  to the  admin.$cmd  collection, i.e.
     // $ admin.$cmd.findOne( { buildinfo: 1 } )
-    procedure BsonWriteParam(const paramDoc: variant);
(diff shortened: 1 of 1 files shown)


synopse (approved):
I guess I will integrate this with a runtime check of the WireProtocol, and allow connection on old and new MongoDB instances.
Checking the protocol version is what is required for a client, IIRC.

Thanks a lot for your input!

danielkuettner:
Hi Arnaud,

thank you for patching my little changes.
I suggest to work with the ifdef and set it to the new protocol version. The old one is never needed and every body should be able to update MongoDB.
It seems that you don’t like Croatia?

Daniel

synopse:
It does make sense to remote the old protocol.

My question with the current implementation is: how do you insert or update items?
The opcodes have been removed, but our driver still use opInsert and opUpdate :(

danielkuettner:
I use only RunCommand, for inserts, updates and deletes too.

Therefore I don’t have touched these parts.

If you need my help…

Daniel

synopse:
And for searches, I guess you use https://www.mongodb.com/docs/manual/reference/command/find/#mongodb-dbcommand-dbcmd.find ?

danielkuettner:
No, RunCommand with aggregate too.

danielkuettner:
You could encapsulate RunCommand for all your other classes (e.g. find, update, insert).

All other code would be obsolete

PR #118, opened 2022-09-11
update allow read from secondary

danielkuettner:
basic setting for using ReadPreference with GetOneReadConnection

### DIFF
--- a/src/db/mormot.db.nosql.mongodb.pas
+++ b/src/db/mormot.db.nosql.mongodb.pas
@@ -2234,6 +2234,10 @@ begin
   else
     fCommand := Command;
   BsonVariantType.AddItem(fCommand, ['$db', fDatabaseName]); // since OP_MSG
+  // since OP_MSG mqfSlaveOk TMongoQueryFlag was replaced by Global Command Argument
+  // https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst
+  BsonVariantType.AddItem(fCommand, ['$readPreference', _OBJ(['mode', 'primaryPreferred'])]);
+
   if not BsonVariantType.IsOfKind(fCommand, betDoc) then
     raise EMongoException.CreateUtf8('%.Create: command?', [self]);
   // generate the proper OP_MSG/OP_COMPRESSED content


synopse:
This leads to a full rewrite of the BSON content, in BsonAddItem().

At least, we could use a single call to AddITem():
```
BsonVariantType.AddItem(fCommand, ['$db', fDatabaseName, 
                                          '$readPreference', _ObjFast(['mode', 'primaryPreferred'])]);
```

But I guess we could include this flag only if necessary, e.g. it is not needed for an insertion.

danielkuettner:
Hi Arnaud,

I’ve just added the second AddItem to show what is missing. I single call is a good idea. In case of insertion it’s not needed but it is just a small extension of the command string and it doesn’t hurt in case of inserting.

Daniel

synopse:
But at command level, if we make a read operation (e.g. find), then it would affect the read performance and only use the primary?

I don't understand how it is expected to work with `TMongoClient.GetOneReadConnection`.
This method is expected to handle the `ReadPreference` property, then pickup the expected connection.

My guess is that the command parameter should match `TMongoClient.ReadPreference` value. 
For instance, the specification states that `For mode 'primary', drivers MUST NOT set $readPreference` so our pull request is not correct in this case.
And how this parameter should apply to the the write operations seems a bit complicated: https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#use-of-read-preferences-with-commands

danielkuettner:
> 
> Am 12.09.2022 um 10:04 schrieb ab ***@***.***>:
> 
> 
> But at command level, if we make a read operation (e.g. find), then it would affect the read performance and only use the primary?
> 
No. Is is only a flag for allowing to read from a secondary. If you read from primary (using a primary ip as mongo client), the option is useless.

> I don't understand how it is expected to work with TMongoClient.GetOneReadConnection.
> This method is expected to handle the ReadPreference property, then pickup the expected connection.
> 
Exactly. And if GetOneReadConnection extracts a secondary (mongo client) and you want to call a find from it, you need this option.

> My guess is that the command parameter should match TMongoClient.ReadPreference value.
> 
The option make only sense in case of using replica sets. This info comes from hello (i guess).
> For instance, the specification states that For mode 'primary', drivers MUST NOT set $readPreference so our pull request is not correct in this case.
> 
I’ve read this too, but my test cases works without errors anyway. So I would do it in such an easy way.
> And how this parameter should apply to the the write operations seems a bit complicated: https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#use-of-read-preferences-with-command <https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#use-of-read-preferences-with-command>If I use RunCommand, I select a connection with GetOneReadConnection. Then I want a result. If I do it wrong (e.g. selecting a secondary for insertion), it’s not an issue of mORMot.

I think the new operator „$readPreference“ is comparable for calling rs.slaveOk() in earlier versions.

danielkuettner:
> 
> Am 12.09.2022 um 10:04 schrieb ab ***@***.***>:
> 
> 
> ... then it would affect the read performance and only use the primary?
> 
I’ve read over this and not answered. What would affect the read performance? The read performance is only affected by the selected member of the replica set. You have always three members, but only primary will be used for all operations. That sounds stupid for me.

Or did you mean, the read performance will be decreased of the second parameter ($readPreference) in case of using not replica sets?. Does your test gives such a result? I didn’t test it, but it would surprising me.

synopse:
About read performance, my guess is that if we set `primaryPreferred` in the command, and it does not match the TMongoClient setting, which may be e.g. `rpSecondary`, then even if we use a secondary connection in `GetOneReadConnection`, it will redirect to the primary node, so make a trip to the secondary for no benefit.

danielkuettner:
Ok, now I understand. But I think it’s impossible.
The client is selected by you. He get the request.
It’s not a mongos.
And anyway, we only set primaryPreferred. I couldn’t see any disadvantages here.

synopse:
BsonVariantType.AddItem() has a real performance problem because of all reallocation.

So please come back with something more tuned, and re-open this issue.

synopse (changes requested):
BsonVariantType.AddItem() has a real performance problem because of all reallocation.

So please come back with something more tuned, and re-open this issue.

danielkuettner:
Ok. How about an overloaded RunCommand method with a param readPreference?

PR #119, opened 2022-09-22
Fixes for ACME client

achechulin:
Some small fixes after refactoring.

About EccGetPubKey. Instead of EC_KEY_key2buf you can use EC_KEY_get0_public_key + EC_KEY_get0_group + EC_POINT_point2buf.


synopse:
Thanks for the feedback!

PR #120, opened 2022-09-29
TAcmeClient fixes

achechulin:
On my Ubuntu 18.04 HEAD request to Directory URL returns empty body. So I changed to GET and now it works.


synopse:
Yes, by definition HEAD returns an empty body (it is a GET without a body).
It was my mistake for sure.

synopse:
Did you make any further tests?

achechulin:
At that point my server successfully received certificates from Let's Encrypt.
Yes, I will test further.

PR #122, opened 2022-10-15
Fix Split when Str is empty

achechulin:
When Str is empty, StartPos  assigned value 0, and last line failed with AV:
`FastSetString(result, @PByteArray(Str)[-1], 1)`


synopse:
Good finding!
Thanks!

PR #125, opened 2022-10-27
Customized Section for INI file

dkounal:
Add the possibility to define the section name when saving in INI file the Tdaemon Settings


synopse (approved):
Good idea.

PR #126, opened 2022-10-30
Minor changes

martin-doyle:
- Reduce threads from 32 down to 4 for this use case
- Make verbose logging available for server daemon


synopse (commented):
Thanks for the feedback!

PR #130, opened 2022-12-01
Fix Websocket over HTTPS

achechulin:
Websocket connection over HTTPS does not work (returns 404) because hsrHttps is missing in Ctxt.ConnectionFlags and TRestHttpServer.Request cannot find REST server instance (in `if (Security in SEC_TLS) = tls then...`). 

This patch changes ComputeContext to include hsrHttps in ConnectionFlags.

Not sure if we need to do this in TWebSocketProcessClient.ComputeContext.


synopse (approved):
Nice finding!
Thanks!

PR #132, opened 2022-12-31
TFB: calc best params depending on CPU count

pavelmash:


synopse (approved):

synopse:
Thanks and a peaceful new year to you! :ukraine:

PR #134, opened 2023-01-10
optimize Postgres bulk update 

pavelmash:
optimize Postgres bulk update  by updating table in primary key order to minimize locks


synopse (approved):
Sounds like a great solution!
Thanks a lot for investigating such a complex problem in Postgres.
Reference post: https://synopse.info/forum/viewtopic.php?pid=38826#p38826

PR #135, opened 2023-01-14
propose to remove SQL statement cache locking 

pavelmash:
because it already thread safe


synopse:
This is unsafe and won't change the performance in case of a single connection for multiple threads.
TSqlDBConnection is not necessary thread-safe.

And for a per-thread TSqlDBConnection, it will never lock, so I won't be a bottleneck.
Better by safe than fooled.

PR #141, opened 2023-01-28
TFB benchmark example synced with latest TFB PR 

pavelmash:
- PG pipelining mode fix (as in TFB PR#7879)
- `pStmt absolute stmt` not work with interface - change back to typecast

### DIFF
--- a/ex/techempower-bench/raw.pas
+++ b/ex/techempower-bench/raw.pas
@@ -15,7 +15,7 @@ See https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-
 // logging is fine for debugging, less for benchmarking ;)
 
 uses
-  {$I mormot.uses.inc} // include mormot.core.fpcx64mm
+  {$I mormot.uses.inc} // include mormot.core.fpcx64mm or mormot.core.fpclibcmm
   sysutils,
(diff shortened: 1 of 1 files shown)


ex/techempower-bench/raw.pas
```diff
@@ -461,23 +461,23 @@ function TRawAsyncServer.rawfortunes(ctxt: THttpServerRequestAbstract): cardinal
   conn: TSqlDBConnection;
   stmt: ISQLDBStatement;
   list: TFortunes;
-  f: TFortune;
   arr: TDynArray;
   n: integer;
+  i: PtrInt;
```
synopse, line 466: you could even write
```
    f: ^TFortune;
...
begin
    f := arr.NewPtr;
    f.id := stmt.ColumnInt(0);
    f.message := stmt.ColumnUtf8(1);
end;
    f := arr.NewPtr;
    f.id := 0;
    f.message := FORTUNES_MESSAGE;
```

synopse (changes requested):

ex/techempower-bench/raw.pas
pavelmash: Thanks. Ready

PR #152, opened 2023-03-02
tfb example: fixed rawupdates

pavelmash:
use ? as params placeholders as expected by mORMot DB layer

Since performance is the same I propose to revert to sub-select version as more realistic


synopse:
Are you sure performance is the same for `/updates?queries=20` on high-end HW ?

synopse (approved):

PR #154, opened 2023-03-07
fix GetFieldData on Delphi UNICODE

cybexr:


synopse (changes requested):
What is the exact problem with the current code?
What is your Delphi version?

I am not even sure what 
```
tmp:= RawByteString( Utf8DecodeToUnicodeString(data, len) );
```
is supposed to do: converting a UnicodeString into RawByteString is not the solution to whatever problem you have.

synopse:
I am confused because this code was as such in mORMot 1 since years with no problem reported:
```
  ftString: begin
    if DataLen<>0 then begin
      CurrentAnsiConvert.UTF8BufferToAnsi(Data,DataLen,Temp);
...
```

synopse:
No news since my last feedback request, one month ago.

Please use the forum for discussion.
I will reopen the ticket when more input is available.

PR #159, opened 2023-03-14
Postgres: bind int4 and int8 parameters as binary

pavelmash:
This MR break parameters logging, since I rewrite a p^.VInt64 value to be in htonl byte order as required by PG binary protocol. 
Can we extend a TSqlDBParam by adding VInt64BE, because comment below say don't to extend....  
```
// - don't change this structure, since it will be serialized as binary
  // for TSqlDBProxyConnectionCommandExecute
  TSqlDBParam = packed record
```

### DIFF
--- a/src/db/mormot.db.sql.postgres.pas
+++ b/src/db/mormot.db.sql.postgres.pas
@@ -232,7 +232,6 @@ type
 implementation
 
 uses
-  sockets, // htonl
   mormot.db.raw.postgres; // raw libpq library API access
 
 
@@ -672,15 +671,14 @@ begin
             begin
               fPGParamFormats[i] := 1; // binary
               fPGParamLengths[i] := 4;
-              p^.VInt64 := htonl(p^.VInt64);
+              p^.VInt64 := bswap32(p^.VInt64);
               fPGParams[i] := @p^.VInt64;
             end
             else if p^.VDBType = INT8OID then
             begin
               fPGParamFormats[i] := 1; // binary
               fPGParamLengths[i] := 8;
-              Int64Rec(p^.VInt64).Hi := htonl(Int64Rec(p^.VInt64).Hi);
-              Int64Rec(p^.VInt64).Lo := htonl(Int64Rec(p^.VInt64).Lo);
+              p^.VInt64 := bswap64(p^.VInt64);
               fPGParams[i] := @p^.VInt64;
             end
             else
--- a/src/db/mormot.db.sql.postgres.pas
+++ b/src/db/mormot.db.sql.postgres.pas
@@ -232,6 +232,7 @@ type
 implementation
 
 uses
+  sockets, // htonl
   mormot.db.raw.postgres; // raw libpq library API access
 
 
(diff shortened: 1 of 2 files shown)


src/db/mormot.db.sql.postgres.pas
```diff
@@ -232,6 +232,7 @@   TSqlDBPostgresStatement = class(TSqlDBStatementWithParamsAndColumns)
 implementation
 
 uses
+  sockets, // htonl
```
synopse, line 235: just use BSwap64() instead from mormot.core.base

src/db/mormot.db.sql.postgres.pas
```diff
@@ -666,7 +667,25 @@ procedure TSqlDBPostgresStatement.BindParams;
...
           p^.VData := '';
         ftInt64:
-          Int64ToUtf8(p^.VInt64, RawUtf8(p^.VData));
+          begin
+            if p^.VDBType = INT4OID then
+            begin
+              fPGParamFormats[i] := 1; // binary
+              fPGParamLengths[i] := 4;
+              p^.VInt64 := htonl(p^.VInt64);
+              fPGParams[i] := @p^.VInt64;
```
synopse, line 676: use bswap32()

src/db/mormot.db.sql.postgres.pas
```diff
@@ -666,7 +667,25 @@ procedure TSqlDBPostgresStatement.BindParams;
...
+            end
+            else if p^.VDBType = INT8OID then
+            begin
+              fPGParamFormats[i] := 1; // binary
+              fPGParamLengths[i] := 8;
+              Int64Rec(p^.VInt64).Hi := htonl(Int64Rec(p^.VInt64).Hi);
+              Int64Rec(p^.VInt64).Lo := htonl(Int64Rec(p^.VInt64).Lo);
+              fPGParams[i] := @p^.VInt64;
```
synopse, line 684: use bsaw64() - and note that you should have switches Hi and Lo for proper work

src/db/mormot.db.sql.postgres.pas
```diff
@@ -725,8 +749,21 @@ procedure TSqlDBPostgresStatement.Prepare(
...
+    // get param types for possible binary binding
+    if fPreparedParamsCount > 0 then
+    begin
+      // allocate params in dynamic array
+      fParam.Count := fPreparedParamsCount;
+      res := PQ.DescribePrepared(c.fPGConn, pointer(fPreparedStmtName));
+      PQ.Check(c.fPGConn, 'DescribePrepared', res, nil, {forceClean=}false);
+      for i := 0 to fPreparedParamsCount - 1 do
+      begin
+        fParams[i].VDBType := PQ.ParamType(res, i);
```
synopse, line 763: couldn't we cache the DB types too ?

synopse (changes requested):
Please see my comments.

Did you see any noticeable performance benefit for the TFB raw benchmark?

synopse:
> This MR break parameters logging, since I rewrite a p^.VInt64 value to be in htonl byte order as required by PG binary protocol. 
One you fixed my comments, I would see how to do that.

pavelmash:
I see a small changes

Bin binds
```
Tests for: raw6_binBins 8 16 16; Tue Mar 14 15:14:13 EET 2023   
rawdb                     474,644  
rawqueries?queries=20      47,017
db                        442,224                 
queries?queries=20         34,254
```
text binds
```
Tests for: raw6_textBinds 8 16 16; Tue Mar 14 15:16:49 EET 2023   
rawdb                     472,289   
rawqueries?queries=20      46,859    
db                        442,889       
queries?queries=20         33,992
```
But I stuck in CPU on my server. I expect bigger difference on TFB hardware

src/db/mormot.db.sql.postgres.pas
pavelmash: done

src/db/mormot.db.sql.postgres.pas
pavelmash: done

synopse:
From what I can see, the log text is computed into `fSqlWithInlinedParams` within `SqlLogBegin(sllSQL)` which is called before `BindParams` so it should have the expected little-endian values.

pavelmash:
Now I see - logging problem appears only for raw*, where  "ExecutePrepared" logged twice. 
 - First by SqlLogBegin(sllSQL);  inside SendPipelinePrepared - here value is OK
 - Second by SqlLogBegin(sllSQL); insige GetPipelineResult - here value is in BE
 
```
16:16:47.256 12 SQL     mormot.db.sql.postgres.TSqlDBPostgresStatement(7f58600010d0) ExecutePrepared t=6us c=01 q=select id,randomNumber from World where id=406

16:16:47.256 12 SQL     mormot.db.sql.postgres.TSqlDBPostgresStatement(7f58600010d0) ExecutePrepared t=73us c=01 r=1 q=select id,randomNumber from World where id=2516647936
```

We should not use SqlLogBegin in second case..... Or let's it be as is....

PR #160, opened 2023-03-14
Changed TServiceCustomStatus to record.

marcosbarreto:
Now to clients no mormot we can receive HttpStatus code as return and body with
out parameters, case status code < HTTP_NOTFOUND
or Exception with custom message when status code >= HTTP_NOTFOUND

This changes fix problem of json return with last array item with statuscode too. some clients cannot recognize this json result

{
    "result": [
        {
            "RowID": 1,
            "Nome": "projeto ab",
            "DataCriacao": 1678706574,
            "UltimaAlteracao": 1678796824,
            "IdCriador": 5,
            "Paineis": []
        },
        0  <<<<<<<<<<<<<<<<<<<<
    ]
}

### DIFF
--- a/src/core/mormot.core.interfaces.pas
+++ b/src/core/mormot.core.interfaces.pas
@@ -503,7 +503,7 @@ type
     // - will handle vtPointer/vtClass/vtObject/vtVariant kind of arguments,
     // appending class name for any class or object, the hexa value for a
     // pointer, or the JSON representation of any supplied TDocVariant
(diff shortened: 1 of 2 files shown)


synopse:
It was as expected that the status is within the JSON.
This code has a meaning even from mORMot clients, and the mORMot clients do not have access to the HTTP status code, when executed from a pascal interface method.

Also raising an exception on server side does not make sense on the server side, because the service could really want to return HTTP codes > 404.

Please use the forum for further discussion, and define what need to be fixed.

PR #163, opened 2023-03-23
Fix TZipRead unzip when DestDir has no end path delimiter

AndreasLrx:
- Unzip('dir/file.txt', 'outdir') was extracting the file at 'outdirdir/file.txt' instead of 'outdir/dir/file.txt'

### DIFF
--- a/src/core/mormot.core.zip.pas
+++ b/src/core/mormot.core.zip.pas
@@ -2813,7 +2813,7 @@ begin
     if not SafeFileName(LocalZipName) then
       raise ESynZip.CreateUtf8('%.UnZip(%): unsafe file name ''%''',
         [self, fFileName, LocalZipName]);
-    Dest := EnsureDirectoryExists(DestDir + ExtractFilePath(LocalZipName));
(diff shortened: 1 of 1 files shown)


src/core/mormot.core.zip.pas
```diff
@@ -2813,7 +2813,7 @@ function TZipRead.UnZip(aIndex: integer; const DestDir: TFileName;
     if not SafeFileName(LocalZipName) then
       raise ESynZip.CreateUtf8('%.UnZip(%): unsafe file name ''%''',
         [self, fFileName, LocalZipName]);
-    Dest := EnsureDirectoryExists(DestDir + ExtractFilePath(LocalZipName));
+    Dest := EnsureDirectoryExists(MakePath([DestDir, ExtractFilePath(LocalZipName)]));
```
synopse, line 2816: My guess is that the correct fix is
`Dest := EnsureDirectoryExists(EnsureDirectoryExists(DestDir) + ExtractFilePath(LocalZipName));`

synopse (changes requested):

synopse (approved):

PR #164, opened 2023-03-23
Update mormot.core.zip.pas

dkounal:


synopse:
Sorry I have fixed it in-between.
:)

PR #168, opened 2023-04-06
add LONGTEXT for mysql and include mysql for test

dkounal:


synopse (approved):
Thanks for the feedback!

PR #169, opened 2023-04-07
Fix Asn1 Boolean encoding

AndreasLrx:


synopse:
see https://github.com/synopse/mORMot2/commit/da76642b

PR #172, opened 2023-04-11
CPU pinning with limited CPUs support 

pavelmash:
- use accessible CPUs (can be limited by taskset) instead of available
- added named command line parameters for servers (-s), threads )-t) and CPU pinning (-p)
- added pinning servers to CPUs
- change rawupdates to use `update table set .. from values (), (), ... where id = id`


pavelmash:
Yes, methods to check available CPUs and command line utils can be moved to
mORMot.

вт, 11 квіт. 2023 р. о 22:30 ab ***@***.***> пише:

PR #176, opened 2023-04-18
Unix-way of formating command line help message

pavelmash:
Before changes
```
raw [options] [params]
  TFB Server using mORMot 2

 [options]
-p --pin:
     pin each server to a CPU
--nopin:
     disable the CPU pinning
-? --help:
     display this message

 [params]
-s --servers <count> (default 12):
     count of servers (listener sockets)
-t --threads <size> (default 8):
     per-server thread pool size
```

After changes

```
TFB Server using mORMot 2

Usage: raw [options] [params]

Options:
  -p, --pin           pin each server to a CPU
      --nopin         disable the CPU pinning
  -?, --help          display this message

Params:
  -s, --servers <count> (default 12)
                      count of servers (listener sockets)
  -t, --threads <size> (default 8)
                      per-server thread pool size
```

### DIFF
--- a/src/core/mormot.core.os.pas
+++ b/src/core/mormot.core.os.pas
@@ -7515,7 +7515,7 @@ procedure TExecutableCommandLine.Describe(const v: array of RawUtf8;
   k: TExecutableCommandLineKind; d, def: RawUtf8; argindex: integer);
 var
   i, j: PtrInt;
-  desc, param: RawUtf8;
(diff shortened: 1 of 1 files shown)


pavelmash:
Columnar align is more easy to read.  May be we should add word-wrap for long descriptions at 80 chars - see `wget --help` for example

PR #179, opened 2023-04-23
 Support for dMariaDB type of database server for zeos

dkounal:
MariaDB supports SQL commands in a way that Mysql does not: ex. IF EXISTS in CREATE INDEX 
MariaDB is now added in zeos 8.0 as an additional protocol.


synopse (approved):
Sounds nice!

Thanks for sharing.

PR #180, opened 2023-04-28
Sync with current TFB sources:

pavelmash:
- removed unneeded Flush in rawqueries;
- added connection distribution output on shutdown
- cached cache table pointer

### DIFF
--- a/ex/techempower-bench/raw.pas
+++ b/ex/techempower-bench/raw.pas
@@ -93,6 +93,7 @@ type
     fModel: TOrmModel;
     fStore: TRestServerDB;
     fTemplate: TSynMustache;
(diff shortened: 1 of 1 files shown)


synopse:
"removed unneeded Flush in rawqueries" is not part of the PR.

synopse (changes requested):
Some missing files I guess.

About asoForceConnectionFlush, putting it sooner is IMHO still a good idea because in GetPipelineResult() it may be too late: this is called in TSqlDBPostgresAsyncThread.Execute within a lock.

pavelmash:
There is no missed files. Can be merged now - it adds
- calculate `fStore.Orm.Cache.Table(TOrmCachedWorld) ` once
- output of connection distribution between servers

PR #181, opened 2023-04-30
decimalseparator is not the dot for all countries

dkounal:
I noticed that the mormot2's original StrToCurrency does not work in case you do not have a '.' as decimalSeparator (In Greece, we have the ',').


synopse:
This is as expected.

This function was not meant for human input conversion, but for internal/computerized JSON-like process.
In your UI, use string and the RTL values using the default locale.
In your business logic, use RawUtf8 and its JSON-oriented functions.
In between, use StringToUtf8() or Utf8ToString().
Or for your use case, use the RTL function to generate a string.

Before creating such a pull request, ask on the forum.

PR #182, opened 2023-05-01
fixed Internal method as published by `FindNonVoid`

flydev-fr:
- In new command-line parser a0d1ec6, `FindNonVoidRawUtf8` & `FindNonVoidRawUtf8I` trigger an access violation when only the switch symbol is given as argument, case-sensitive or not.

### DIFF
--- a/src/core/mormot.core.base.pas
+++ b/src/core/mormot.core.base.pas
@@ -4701,7 +4701,7 @@ begin
   result := 0;
   repeat
     p := n[result]; // all VName[]<>'' so p=n^<>nil
-    if (PStrLen(p - _STRLEN)^ = len) and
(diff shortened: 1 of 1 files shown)


synopse:
This is not the proper fix.
You change the behavior of a function which is named `FindNonVoidRawUtf8`
The caller should be fixed, because the function is EXPECTED to have non-void values.

Please use the forum for such discussion.

synopse:
I suspect we did make a fix on last Friday, which may be related to your issue.
https://github.com/synopse/mORMot2/commit/554ff23ae0bc743003ac137fd7525f074166d01c

Please try with the latest version of the framework.

synopse:
After investigation, it seems that last week fix was not enough.

Please try with https://github.com/synopse/mORMot2/commit/38f27f08

PR #194, opened 2023-06-02
Fix infinite loop when using multiple times the same parameter

AndreasLrx:
With a command line like this:
```sh
binary --parameter 0 --parameter 1
```

A call to `Get(['parameter'], utf8DynArray)` was creating an infinite loop


synopse (approved):
nice finding!

PR #201, opened 2023-06-12
Update mormot.lib.curl.pas

danielkuettner:
added missing Type for file upload callback


synopse:
see https://github.com/synopse/mORMot2/commit/4a372146

PR #224, opened 2023-10-21
improve IPToCardinal doc to avoid confusion

pavelmash:
While try to use this function in mORMot1 my first attempt was invalid, because I'm not expect  what result is in "reverse" order. Better to note this in documentation for mROMot2


synopse (commented):
Instead of "reverse" order, please write "little endian order, not network order".
The "big endian" order is the reverse to the "little endian". And the opposite is also true. ;)

pavelmash:
Done

PR #225, opened 2023-10-23
Preallocation for text output parameters for Oracle SQL statements.

ASiwon:
Changes according to the mORMot forums topic: https://synopse.info/forum/viewtopic.php?id=6402 for mORMot 2.

### DIFF
--- a/src/core/mormot.core.base.pas
+++ b/src/core/mormot.core.base.pas
@@ -110,6 +110,9 @@ const
   /// use rather CP_LATIN1 with mORMot 2
   CODEPAGE_LATIN1 = CP_LATIN1;
 
+  /// Maximum bytes for single UTF-8 char.
+  UTF8_CHAR_MAX_BYTES = 4;
+
 {$ifdef FPC} { make cross-compiler and cross-CPU types available to Delphi }
 
 type
--- a/src/db/mormot.db.raw.oracle.pas
+++ b/src/db/mormot.db.raw.oracle.pas
@@ -897,6 +897,8 @@ type
   public
     // the client verion numbers
     major_version, minor_version, update_num, patch_num, port_update_num: sword;
+    /// Maximum supported varchar string length
+    MaxVarCharLength: Integer;
     /// if OCI handles directly Int64 bound parameters (revision >= 11.2)
     SupportsInt64Params: boolean;
     /// OCI will call OCILobGetChunkSize when retrieving BLOB/CLOB content
@@ -1693,6 +1695,10 @@ begin
   SupportsInt64Params := (major_version > 11) or
                          ((major_version = 11) and
                           (minor_version > 1));
+  if major_version > 7 then
+    MaxVarCharLength := 4000
+  else
+    MaxVarCharLength := 2000;
   UseLobChunks := true; // by default
 end;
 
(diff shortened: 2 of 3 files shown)


src/core/mormot.core.base.pas
```diff
@@ -110,6 +110,9 @@ interface
   /// use rather CP_LATIN1 with mORMot 2
   CODEPAGE_LATIN1 = CP_LATIN1;
 
+  /// Maximum bytes for single UTF-8 char.
```
synopse, line 113: It is not the right place to put this.
And it is a very difficult guess to make, because we support up to 6 bytes per Unicode point, which theoretically could be 32-bit.
Note that there is no such this as an "UTF-8 char".
So I would not define such a constant.

src/db/mormot.db.raw.oracle.pas
```diff
@@ -1693,6 +1695,10 @@ constructor TSqlDBOracleLib.Create(LibraryFileName: TFileName);
   SupportsInt64Params := (major_version > 11) or
                          ((major_version = 11) and
                           (minor_version > 1));
+  if major_version > 7 then
```
synopse, line 1698: Why do we need to change according to the version?

src/db/mormot.db.raw.oracle.pas
```diff
@@ -897,6 +897,8 @@   TSqlDBOracleLib = class(TSynLibrary)
   public
     // the client verion numbers
     major_version, minor_version, update_num, patch_num, port_update_num: sword;
+    /// Maximum supported varchar string length
```
synopse, line 900: Do not use "varchar string length", but just "varchar bytes".
My guess is that it should be part of the oracle driver properties, not at the OracleLib level.

synopse (changes requested):
Thanks for the feedback.

Please see my comments.

src/db/mormot.db.raw.oracle.pas
ASiwon: Change according to the version is not a very important thing. Newer Oracle versions accepts longer varchar param so code for newer version will work also for older Oracle version. At most more memory will be allocated for buffers, but this should not have negative consequences, because memory will be released just after SQL execution.

src/db/mormot.db.raw.oracle.pas
ASiwon: Ok.

src/core/mormot.core.base.pas
ASiwon: Ok. I will change this.

ASiwon:
Yesterday I checked the Oracle documentation and found some new information about varchar parameter's length rules. I think it will be possible to base directly on lengths defined in bytes and resign on lengths defined in chars. I must test it in my Oracle servers so please be a patient.

synopse:
> Yesterday I checked the Oracle documentation and found some new information about varchar parameter's length rules. I think it will be possible to base directly on lengths defined in bytes and resign on lengths defined in chars. I must test it in my Oracle servers so please be a patient.

Great!
We are looking forward to this.

synopse:
Any update on your side?

synopse:
Any news?

ASiwon:
My apologies. Last weeks I was totally busy in my job and had no time make this correction. I will try to make the changes in this weekend.

synopse:
Closing due to no feedback.

Please reopen with new material.
Thanks!

PR #226, opened 2023-10-30
Pass optional UserName&Password params to Service Install

the-real-wizard-of-oz:
Small change, default const params for Username and Password provide full backward-compatibility.


synopse:
Simple and may be useful.
Thanks for sharing!

PR #242, opened 2024-03-01
RFC 9239

sakura1977:
Two changes according to RFC 9239

* "text/javascript" as mime content type for JS-files, "application/javascript" is defined as obsolete
* MJS-filetype to be handled like JS-filetype

https://datatracker.ietf.org/doc/html/rfc9239
https://www.rfc-editor.org/rfc/rfc9239

### DIFF
--- a/src/core/mormot.core.buffers.pas
+++ b/src/core/mormot.core.buffers.pas
@@ -8681,20 +8681,20 @@ begin
 end;
 
 const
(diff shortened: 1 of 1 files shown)


synopse (approved):

synopse:
Thanks for the input!

synopse:
Needed some minor fixes.

PR #243, opened 2024-03-17
SimpleClient bug-fix for pas2js and other clients. M1->M2 bugfix

TTomas:
Comments in code.

### DIFF
--- a/src/soa/mormot.soa.codegen.pas
+++ b/src/soa/mormot.soa.codegen.pas
@@ -592,6 +592,7 @@ type
     fSOA: variant;
     fSourcePath: TFileNameDynArray;
     fHasAnyRecord: boolean;
+    fNestedId: Integer;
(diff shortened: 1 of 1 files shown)


synopse:
Thanks a lot.

See my comment above.

TTomas:
About nested type names, in ContextFromRtti this code works great for me. Add fNestedId: integer in Class protected, needed if different records/arrays have same nested field name.
[code]
  // set typName/typAsName
  if typName = '' then
    if rtti <> nil then
    begin
+     if rcfWithoutRtti in rtti.Flags then
+       typName := 'T'+parentName+IntToStr(InterlockedIncrement(fNestedId))
+     else
        typName := rtti.Name
    end
    else
      typName := TYPES_LANG[lngDelphi, typ];
[/code]

synopse:
I will now review the code after merging.

Thanks a lot for all this debugging and sharing !

PR #267, opened 2024-07-08
Fix UrlDecodeInt64

Coldzer0:
The changes in this commit https://github.com/synopse/mORMot2/commit/c27ec5dc547f9ef0b48e3b24aeaa285051f5afd1

Introduced a bug in UrlDecodeInt64, which returns the opposite sign of a value.


Coldzer0:
I suggest adding some test cases for these functions to ensure that any future changes will not break them.

synopse:
Thanks for the feedback.

I have included some regression tests.

PR #276, opened 2024-08-01
mormot: reintroduced dropped DrawAntiAliased() for mormot2

dhewg:
See a455fd3e "SynGdiPlus: add support for image attributes" on the v1 repo.


synopse (approved):
Some minor tweaks, I will fix during merging.

Thanks for the idea.

PR #277, opened 2024-08-01
mormot: fix off-by-one in FindNameValue()

dhewg:
This patch is a bit dated and I forgot the details how I ran into the issue :\
It probably was a tab?
Hopefully the fix is obvious to the reviewer ;)


synopse:
Sounds fine to me.

PR #279, opened 2024-08-05
fix: typo in mvc-blog example

flydev-fr:


synopse (approved):

synopse:
Thanks for the feedback!

PR #281, opened 2024-08-12
Introduced new overloaded function RunProcess().

the-real-wizard-of-oz:


synopse:
My guess is that it won't compile on POSIX/non Windows, because both overloads would have the same signature.

synopse:
And also PProcessInformation vs PHandle is very confusing. I suspect FPC or some Delphi would have troubles with it.

So, perhaps not an overload, but a new Windows-only method like `RunCommandWin()` returning PProcessInformation.

the-real-wizard-of-oz:
Current commit compiles well on FPC. Should i rename RunProcess() to RunProcessWin() anyway?

synopse:
Please see my last commit.
https://github.com/synopse/mORMot2/commit/5212b264

synopse:
Closing for now, hoping my last commit fits your needs.
:)

Please reopen if it was not enough.
Thanks.

the-real-wizard-of-oz:
RunProcessWin() seems fine, thanks.

PR #282, opened 2024-08-12
Include Dicom to mime types

dkounal:
I have added the dicom files' mime type. 
The Magic bytes for the file type is from here: https://dicom.nema.org/medical/dicom/current/output/chtml/part10/chapter_7.html
I have included the usually used file extensions: .DCM & .DICOM The extension .DIC was used more in the past but this extension can be used for other types of files and I did not included it. 
I am not sure about the position of new file extensions in MIME_EXT. Please check it and corrected it if needed.


synopse:
Thanks!

PR #285, opened 2024-08-26
Bugfix: lacking a begin..end block

JavierusTk:
The indentation hints the else block should be one statement, but there are two. Compiling in Delphi 12 it shows a hint there


synopse (approved):
Thanks a lot!

PR #286, opened 2024-08-26
WIP - IJsonClient generator from OpenApi specifications file

AndreasLrx:
Introduce a new mormot.net.openapi.pas unit to generate Fpc units from [OpenApi specifications](https://swagger.io/specification/) files.

For the moment, this is able to generate two units from a json file: A file for the DTOs (enums, records, array of records...) and another for an ApiClient (using the new IJsonClient).
This allows to call the Api endpoints as FPC functions with in/out serialization using the RTTIs.

There is still a lot of documentation to write and tests to do with an actual server but I open this PR in case you have some comments of what I've done so far.

And the memory management to handle way better because, I know, nothing is free at the moment.

### DIFF
--- a/src/net/mormot.net.openapi.pas
+++ b/src/net/mormot.net.openapi.pas
@@ -10,7 +10,8 @@ uses
   mormot.core.base,
   mormot.core.variants,
   mormot.core.text,
-  mormot.core.collections;
+  mormot.core.collections,
+  mormot.rest.core;
 
 const
   SCHEMA_TYPE_ARRAY: RawUtf8 = 'array';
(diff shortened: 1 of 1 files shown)


synopse (approved):
Thanks a lot.
This is an exciting new feature.

What did you test yet?

I will merge it but then quickly change it, because your .lpk seems broken, and there is a lot to do to make it following the mORMot requirements (e.g. Delphi compatibility).

AndreasLrx:
Since I first made this as a python script, I tested with the code generated from the python script.

For the DTO unit file, I checked if there was any difference between the python/mORMot generated files and there wasn't.
However for the Client unit file, I changed to use the new `IJsonClient` so I couldn't compare the files. And I don't have a server to test for now.

But I still have fixes to make, that's why I opened this PR in draft and WIP.
I will continue to work on it at least until Wednesday ;)

synopse:
First Pass. ;)

synopse:
There are overlapping GetSchema functions or fSchema properties in TPascalCustomType and its descendant.
It is a bit rude.

synopse:
Second pass:
- compatibility with Delphi down to the very old Delphi 7.
- a lot of refactoring.
- merging your #287 pull request.

Now I add some unit tests.

PR #287, opened 2024-08-26
OpenApi new ExportToDirectory method and fixes

AndreasLrx:
Fixed the memory leaks and some inattention errors.
Don't know if my usage of IJsonClient is valid but at least the generated units compiles and the Rtti registers in the initialization does not raise any errors.

I still have to handle array of array types and default values in the method parameters but since you're also working on this unit I prefer to do small PRs ;)

### DIFF
--- a/src/net/mormot.net.openapi.pas
+++ b/src/net/mormot.net.openapi.pas
@@ -114,92 +114,95 @@ type
   /// a dynamic array of pointers wrapper to OpenAPI Parameter(s)
   POpenApiParameterDynArray = array of POpenApiParameter;
 
-  TOpenApiOperation = record
-  private
-    function GetDeprecated: Boolean;
-    function GetDescription: RawUtf8;
-    function GetDV: PDocVariantData;
-    function GetId: RawUtf8;
-    function GetParameterByIndex(aIndex: Integer): POpenApiParameter;
-    function GetParameters: PDocVariantData;
-    function GetPayloadParameter: POpenApiParameter;
-    function GetResponseForStatusCode(aStatusCode: Integer): POpenApiResponse;
-    function GetResponses: PDocVariantData;
-    function GetSummary: RawUtf8;
-    function GetTags: PDocVariantData;
-  public
-    property DocVariant: PDocVariantData read GetDV;
-
-    property Id: RawUtf8 read GetId;
-    property Summary: RawUtf8 read GetSummary;
-    property Description: RawUtf8 read GetDescription;
-    property Tags: PDocVariantData read GetTags;
-    property Deprecated: Boolean read GetDeprecated;
-
-    property Parameters: PDocVariantData read GetParameters;
-    property Parameter[aIndex: Integer]: POpenApiParameter read GetParameterByIndex;
-    property PayloadParameter: POpenApiParameter read GetPayloadParameter;
-
-    property Responses: PDocVariantData read GetResponses;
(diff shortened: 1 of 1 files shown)


AndreasLrx:
Btw, I'm continuing the refactor on the other records following the first one you refactored in b0f1d65

synopse:
Since I am also working on my side, it indeed makes a lot of conflicts.

synopse:
I tried from
https://raw.githubusercontent.com/OAI/OpenAPI-Specification/main/examples/v3.0/petstore.json

A lot more work to do and fix, I guess.

synopse:
Another pass with
https://github.com/synopse/mORMot2/commit/e3b89bfd
Some JSON do not have any "tags" attached.
And other fixes/enhancements.

I would like to validate it with some reference material from OpenAPI as 2.0 and 3.0 at least.

We still have to find a way to manage errors.
Do you think a dedicated exception class could be a good implementation?
My guess is that it would be what most end users expects. And we could put the error properties into the exception class.
What do you think?

AndreasLrx:
Thanks for your work, that's way better now.

For the version 3.0 it will mostly be refactoring with the way we're dealing references since in 2.0 it is `definition/SchemaName` but in 3.0 it is `components/type/ObjectName` and it can contains schemas, response, parameters... 
But it is mostly some conditions in the TOpenApiSpecs functions and TOpenApiParser reference resolution.

For the errors I assume you mean for http client error responses (4xx) with attached schema in the specifications.
If that's the case, I agree generating records for the response types and then exception class with the record as a `error` attribute seems the way to implement it.

And since we're talking about errors, do you think we should implement parameters validation on client side ? (using properties like minimum, maximum, minimumLength, pattern...) I feel like this is more a server side implementation but it can be a parameter of the generator, if that can help reduce server load it can sometimes be worth it.

Lastly I know there is an issue with the record ordering function: It cause an infinite loop for nested records. 
Because yes it is possible to have an optional field of type A in a record A in the openapi specs (using $ref). But we cannot implement it in Pascal.
I thought of replacing the nested field type with a TDocVariantData, and add a function to convert the field to the record, since the rtti definition are already defined.
Or we just don't handle it and raise an error if a max recursion depth is reached.

Anyway, this generator looks very exciting to use, hope it will bring new mORMot users ;)

synopse:
A lot of work is due to restrictions on Delphi 7. For instance, no generics, no 'for in' statement... It is a bit more verbose, but we have some structures in mORMot - like TRawUtf8List - which help.

I will look into basic 3.0 support, at least for some of the reference material supplied on OpenAPI github.
Anyway, I suspect Swagger 2.0 format is still the most used and generated in the wild.

For errors, yes, it was about translating 4xx responses into exceptions, not client-side validation.

I am not convinced about the use of `base: TReferencedType;` in records.
```

type

  TNewPet = packed record
    Name: RawUtf8;
    Tag: RawUtf8;
  end;

  TPet = packed record
    Base: TNewPet;
    Id: Int64;
  end;
  TPetDynArray = array of TPet;

  TErrorModel = packed record
    Code: integer;
    Message: RawUtf8;
  end;
  ```
Perhaps putting directly the field could be better. This is clearly where we miss `object()` inheritance in modern Delphi. We could use `TPet = object(TNewPet)` on FPC, but `object()` is sadly broken on modern Delphi (since Delphi 2010) so we can't use it in our code. 
My proposal may be to just put the $refs fields into the record, as it does appear in the JSON anyway.

PR #289, opened 2024-08-30
Handle simple OpenApi V3 specifications

AndreasLrx:
Only supports references to schemas, not to responses or other types


synopse (approved):
Very nice!

PR #294, opened 2024-09-05
Robust TOrmModel.Destroy, in case TableProps[i] = nil

michaliskambi:
In case you declare an ORM class with `Single` property, like this:

```delphi
  TOrmCastleTransform = class(TOrm)
    FTranslationX: Single;
  published
    property TranslationX: Single read FTranslationX write FTranslationX;
  end;
```

the constructor of `TOrmModel` will fail with a helpful exception that `oftFloat/rkFloat` are not supported. (It would be even better if the exception message explicitly said that `Single` is not supported but `Double` is :) ).

However (if you run without debugger) it is easy to miss, because it is then obscured by another exception: an access violation occurs when destructor tries to access `TableProps[i].Props`, but `TableProps[i]` is `nil`.

This PR makes the destructor "robust", it will continue without causing a new exception when `TableProps[i] = nil`.

### DIFF
--- a/src/orm/mormot.orm.core.pas
+++ b/src/orm/mormot.orm.core.pas
@@ -10356,23 +10356,30 @@ var
   i, j: PtrInt;
 begin
   for i := 0 to fTablesMax do
-    with TableProps[i].Props do
-    begin
-      fSafe.Lock; // may be called from several threads at once
-      try
-        for j := 0 to fModelMax do
-          if fModel[j].Model = self then
-          begin
-            // un-associate this TOrm with this model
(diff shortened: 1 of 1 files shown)


michaliskambi:
For reference, this is what the developer will see (when running the application without debugger) when trying to use `Single` in ORM class before this PR:

Zrzut ekranu 2024-09-05 033807

And this is after -- looks more obvious what's wrong:

Zrzut ekranu 2024-09-05 034607

synopse (approved):
A bit too verbose, but nice catch.

PR #307, opened 2024-10-22
fix CldapGetDomainInfo: NTVER filter ntver=\06\00\00\00 now works wit…

htouvet:
…h reworked RawLdapTranslateFilter


synopse:
Thanks a lot Hübert for the PR !

PR #310, opened 2024-10-26
Create restws_SimpleEchoServer.dpr - sample taken from mORMot1 and updated

JavierusTk:
Updated from same example in mORMot 1


synopse:
Sounds just fine.
I have made some minor refactoring (naming + formatting + missing .html file).

Thanks for the input!

PR #311, opened 2024-10-27
TSimpleHttpClient.Create() 

zen010101:
add two new params to the Create constructor with default value : aCreateTimoutMS: integer = 0; aIgnoreCertificateErrors: boolean = false


synopse (changes requested):
There seem to be more code than expected in your pull request.

PR #314, opened 2024-11-04
net: Fixed OnWSClose() sometimes not being called normally

zen010101:
- Ensure that the OnWSClose() is called regardless of whether the client connection is disconnected normally or abnormally


synopse:
Thanks for the input.

Why not just add `Assigned(OnWSAsyncClose) or` in the initial code?

zen010101:
OnWSAsyncClose() is a method or a variable. `Assigned(OnWSAyncClose)` will cause the compile error:
```
mormot.rest.http.server.pas(1398,36) Error: Incompatible types: got "TRestHttpServer.OnWSAsyncClose(TWebSocketAsyncConnection);" expected "Int64"
```

synopse (approved):
Make sense!

Thanks for the input!

PR #315, opened 2024-11-04
fix: call `inherited Create` in THttpProxyServer.Create()

zen010101:


synopse (approved):
Thanks for the feedback.

As discussed in https://synopse.info/forum/viewtopic.php?id=7054

PR #316, opened 2024-11-05
ex: fixed '/' lost in settingsfolder

zen010101:


synopse:
The natural way is to call `IncludeTrailingPathDelimiter(settingsfolder) `.

synopse (approved):

PR #317, opened 2024-11-08
Add HTTP-PATCH support to TUriRouter

iret-de:


synopse (approved):
Seems just fine.
Thanks a lot!

PR #319, opened 2024-11-11
fix: memory leak

zen010101:


synopse (approved):
Seems indeed better like that. 
:)

PR #334, opened 2025-01-17
curl: add missing blob constants and type

dhewg:
These are just missing, taken from curl's easy.h

Tested successfully with:
```
blob: TCurlBlob;
ca: TExecutableResource;

FillCharFast(blob, SizeOf(blob), 0);
if ca.Open('cabundle', RT_RCDATA, hInstance) then
begin
  blob.data:= ca.Buffer;
  blob.len:= ca.Size;
  blob.flags:= CURL_BLOB_NOCOPY;
end;
curl.easy_setopt(handle, coCAInfoBlob, @blob);
```


synopse:
Thanks a lot for your input.

PR #337, opened 2025-01-26
fix: Utf8ToConsole, Added a judgment on whether it is necessary to co…

zen010101:
…nvert UTF8 to CP_OEM


synopse:
Your code would work only on FPC.
We need to have a Delphi compatible code too.
I will look into it.

PR #348, opened 2025-03-26
Fix compiling error when NOEXCEPTIONINTERCEPT is used

Coldzer0:
Hello @synopse 

The latest update to the core.log has some forgotten unwrapped code for NOEXC

like

https://github.com/synopse/mORMot2/blob/dd4376c74344820c20567a2acfa2b8a6f07407e3/src/core/mormot.core.log.pas#L4161

https://github.com/synopse/mORMot2/blob/dd4376c74344820c20567a2acfa2b8a6f07407e3/src/core/mormot.core.log.pas#L4619

### DIFF
--- a/src/core/mormot.core.log.pas
+++ b/src/core/mormot.core.log.pas
@@ -4616,7 +4616,9 @@ begin
       exit;
     // reset the current thread context
     nfo^.Index := 0;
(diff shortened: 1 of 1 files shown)


synopse:
You are right.
Sorry for the issue.

PR #354, opened 2025-04-11
ex: fix MVC sample compilation with PUREMORMOT2 defined

flydev-fr:


synopse (approved):

synopse:
Thanks for the input!

PR #356, opened 2025-04-14
ldap: Fixed delete children in non-leaf objects

osvegn:
This Pull Requests contains modification to mormot.net.ldap. It fixes the delete method of TLdapClient. With the deleteChildren parameter to true, children of an object wasn't retrieved due to missplaced argument in search method (placed in filter parameter instead of baseDN). Fixed by adding an empty filter, then the baseDN.


synopse (approved):
Nice catch!

Thanks for sharing.

PR #357, opened 2025-04-16
Fix declaration of GetJpegSize() which is called in a wrong parameter order

EchseKuhTiere:
This fix restores the PDF image embedding which is currently called in a wrong parameter order where `width` and `height` are switched.

In `TPdfImage.CreateJpegDirect()`:
```pas
if not GetJpegSize(aJpegFile.Memory, len, fPixelWidth, fPixelHeight, bits) then
```

but the function is declared as:
```pas
function GetJpegSize(jpeg: PAnsiChar; len: PtrInt; out Height, Width, Bits: integer)
```

The function declaration should be fixed as the order in the old mormot `SynPDF` unit was also `width` before `height` which also is the typical order of those parameters.


synopse (approved):
Make sense!

Thanks for sharing!

PR #359, opened 2025-05-06
New overloaded function for RunCommandWin() with job THandle param.

the-real-wizard-of-oz:
Hi Arnaud, we need job THandle of RunCommandWin() as external param.


synopse:
Should be fine with my previous commit above.

Closing from now - please reopen or discuss on our forum for more.

PR #371, opened 2025-08-01
Support for zerossl.com ACME server

achechulin:
To support zerossl.com ACME server we need to fix two issues:

First, zerossl.com require waiting for certificate. So we add new `FinalizeDomainRegistration` and wait loop on `CompleteDomainRegistration` after call to `FinalizeDomainRegistration` in `RegisterAndWait`.

Second, we add `externalAccountBinding` support to associate an ACME account with an existing account in zerossl.com.

These changes are split into two commit to simplify code review.

Discussion here:
https://synopse.info/forum/viewtopic.php?pid=44091

### DIFF
--- a/src/net/mormot.net.acme.pas
+++ b/src/net/mormot.net.acme.pas
@@ -149,6 +149,7 @@ type
     fNewAccount: RawUtf8;
     fNewOrder: RawUtf8;
     // URI filled by CreateOrder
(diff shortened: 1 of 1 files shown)


synopse:
Very nice!

I suspect the code still works well with Let's Encrypt? ;)

PR #372, opened 2025-08-05
Some fixes to OpenSSL and ACME

achechulin:
First commit fixed X509 certificate refcounting in TOpenSslNetTls.SetupCtx

From OpenSSL docs for SSL_CTX_use_certificate (https://docs.openssl.org/master/man3/SSL_CTX_use_certificate/):

> On success the reference counter of the x is incremented.

From OpenSSL docs for SSL_CTX_add_extra_chain_cert (https://docs.openssl.org/master/man3/SSL_CTX_add_extra_chain_cert/):

> The x509 certificate provided to SSL_CTX_add_extra_chain_cert() will be freed by the library when the SSL_CTX is destroyed. An application should not free the x509 object.

Second commit fixes certificate files renaming.

On first run there is no cert files, DeleteFile returns false, and files are not moved.


synopse:
For DeleteFile() you are obviously correct.

About the reference counting, I am not sure we need to manually handle them as such.
AFAICT `PX509DynArrayFree(xa);` does not free xa[], it actually decrement each refcnt, which is perfectly fine with our pattern.

Line 10972, it is even worse, because you increment manually a refcnt which is already incremented by SSL_CTX_add_extra_chain_cert().

achechulin:
In current implementation there is an error in `SSL_accept` with message "x509 certificate routines::passed a null parameter".

> because you increment manually a refcnt which is already incremented by SSL_CTX_add_extra_chain_cert

`SSL_CTX_add_extra_chain_cert` does not increment refcount.

We can see in source code, that only `sk_X509_push` called, and `sk_X509_push` does not change refcount:
https://github.com/openssl/openssl/blob/openssl-3.0/ssl/s3_lib.c#L3958

Tests also used `X509_up_ref` + `SSL_CTX_add_extra_chain_cert`:
https://github.com/openssl/openssl/blob/openssl-3.0/test/sslapitest.c#L981

And description of OpenSSL stacks and example:
https://man.openbsd.org/STACK_OF.3

PR #395, opened 2025-10-22
ldap: clear `TLdapClient.fNetbiosDN` on reset connection

osvegn:
On changing the TLdapClient instance connection, the fNetBiosDN from the first instance is keeped in TLdapClient instance. It should be removed to ensure the correct fNetBiosDN is retrieve in the Active Directory


synopse:
Thanks for the feedback.

Shouldn't it be part of the "from Close: full reset" begin/end instead?

synopse (approved):

PR #404, opened 2025-10-31
ldap - Support MsDSSupportedEncryptionTypes flags

osvegn:


synopse:
Seems fine!

Thanks!

osvegn:
Did I miss something with this pull request ? It has been approved and closed without merged commits :/

synopse:
I will change it anyway,  because we could do it much easier since each bit in the set match the binary level (no need of constants).

PR #434, opened 2026-02-05
Samples updated

martin-doyle:
Some changes to the user interface


ex/ThirdPartyDemos/martin-doyle/10-InvoiceExample/Data/Project10.db
synopse: Please don't put binaries in the repo.

synopse (changes requested):
Please remove the binary DB file.

synopse:
Perhaps worth a discussion on our forum or Delphi Praxies. :+1: 
Your feedback is amazing!

I will merge the content.

PR #466, opened 2026-04-21
feat: YAML 1.2 parser + OpenAPI YAML spec dispatch

zen010101:
## Summary

- Adds `mormot.core.yaml` — YAML 1.2 core-schema subset, bidirectional
  TDocVariant conversion (parser YAML -> JSON -> InitJson, writer via
  TTextWriter), guarded by `YamlMaxDepth` to turn would-be EStackOverflow
  on deeply nested specs into a clean `EYamlException`.
- Routes `TOpenApiParser.ParseFile` by file extension (`.yaml`/`.yml` ->
  YAML path, otherwise JSON); adds `ParseYaml` / `ParseYamlFile` symmetric
  entry points (mirror of `ParseJson`).
- `TPascalException.Create` accepts `obtVariant` as a generic
  `VariantResponse` fallback — empty-object response schemas (e.g. GitHub
  REST "304 Not Modified") were previously rejected with "unsupported
  schema".

## Test plan

- [x] `TTestCoreYaml`: 28 goldens + 9 error cases + `_Roundtrip` +
  `_FileApi` + `_RecursionDepth` + `_OpenApiEquivalence` —
  `0 / 179` assertions pass.
- [x] `TNetworkProtocols._OpenApiYamlDispatch`: `ParseYaml` /
  `ParseYamlFile` / `ParseFile(.yaml)` produce output byte-identical to
  `ParseJson` for the same spec.
- [x] Full regression on FPC 3.2.3 i386-win32: all assertions pass (one
  pre-existing host-dependent `Dmi smbios` OEM check reproduces with
  `--test TTestCoreBase` on a clean upstream checkout and is unrelated
  to this PR).
- [x] Stress-tested against four public OpenAPI specs:
  - Petstore v3 (22 KB -> 9 KB / 321 lines client)
  - DigitalOcean public API (100 KB -> 135 KB / 3660 lines)
  - Stripe spec3 (6 MB -> 1.25 MB / 27064 lines)
  - GitHub REST API (9.2 MB -> 2.44 MB / 62001 lines)

  All four pass end-to-end through `mopenapi`.

## Explicitly out of scope

- `&anchor` / `*alias` / `!!tag` / multi-document `---` / tab
  indentation / `%YAML` `%TAG` directives — raise `EYamlException` with
  1-based line number (documented in the unit header).
- Two independent `mopenapi` codegen bugs surfaced by compile-checking
  the Stripe and GitHub generated clients (record parameter passed to
  `array of const`; `+1`/`-1` schema field names sanitized to a bare
  digit) — to be filed as separate issues.
- A small follow-up fix for `TPascalEnum.Create` enum-prefix collisions
  beyond 99 anonymous enums (Stripe / GitHub each have around 100
  anonymous enums sharing the same 2-letter initials, which exhausts
  the current `ES2..ES100` suffix range) — to be submitted as a separate
  PR.

### DIFF
--- a/src/core/mormot.core.yaml.pas
+++ b/src/core/mormot.core.yaml.pas
@@ -132,20 +132,15 @@ end;
 // - accepts both LF and CRLF terminators; trailing \r is stripped
 procedure SplitYamlLines(const Yaml: RawUtf8; out Lines: TYamlLines);
 var
-  p, lineStart, eol: PUtf8Char;
-  stop: PUtf8Char;
-  idx, len: integer;
+  p, lineStart, eol, stop: PUtf8Char;
+  idx, len: PtrInt;
   raw: RawUtf8;
+  current: ^TYamlLine;
 begin
-  Lines := nil;
-  SetLength(Lines, 16);
-  idx := 0;
   p := pointer(Yaml);
   if p = nil then
-  begin
-    Lines := nil;
     exit;
-  end;
+  idx := 0;
   stop := p + length(Yaml);
   lineStart := p;
   while p <= stop do
@@ -159,15 +154,16 @@ begin
       else
         FastSetString(raw, lineStart, eol - lineStart);
       if idx >= length(Lines) then
-        SetLength(Lines, length(Lines) * 2);
-      Lines[idx].Raw := raw;
-      Lines[idx].Indent := CountLeadingSpaces(raw);
-      len := length(raw) - Lines[idx].Indent;
+        SetLength(Lines, NextGrow(idx));
+      current := @Lines[idx];
+      current^.Raw := raw;
+      current^.Indent := CountLeadingSpaces(raw);
+      len := length(raw) - current^.Indent;
       // also right-trim trailing spaces from Content
-      while (len > 0) and (raw[Lines[idx].Indent + len] = ' ') do
+      while (len > 0) and (raw[current^.Indent + len] = ' ') do
         dec(len);
-      FastSetString(Lines[idx].Content,
-        pointer(PtrUInt(pointer(raw)) + PtrUInt(Lines[idx].Indent)), len);
+      FastSetString(current^.Content,
+        @PByteArray(pointer(raw))[current^.Indent], len);
       inc(idx);
       if p = stop then
         break;
@@ -175,7 +171,8 @@ begin
     end;
     inc(p);
   end;
-  SetLength(Lines, idx);
+  if length(Lines) <> idx then
+    SetLength(Lines, idx);
 end;
 
 // strip an unquoted trailing "# ..." comment from a scalar fragment
@@ -420,7 +417,6 @@ var
   p, stop: PUtf8Char;
   quote: AnsiChar;
   tmp: TTextWriter;
-  wbuf: TTextWriterStackBuffer;
   ch: AnsiChar;
   c: cardinal;
   i: PtrInt;
@@ -437,7 +433,7 @@ begin
   if quote = '''' then
   begin
     // single-quoted: only '' is an escape
-    tmp := TTextWriter.CreateOwnedStream(wbuf);
+    tmp := TTextWriter.CreateOwnedStream;
     try
       inc(p);
       while p < stop do
@@ -464,7 +460,7 @@ begin
     exit; // unclosed
   end;
   // double-quoted: full escape set
-  tmp := TTextWriter.CreateOwnedStream(wbuf);
+  tmp := TTextWriter.CreateOwnedStream;
   try
     inc(p);
     while p < stop do
@@ -594,7 +590,6 @@ type
     fDepth: integer;
     fMaxDepth: integer;
     fOut: TJsonWriter;
-    fOutBuf: TTextWriterStackBuffer;
     procedure Error(LineIdx: integer; const Msg: RawUtf8); overload;
     procedure ErrorFmt(LineIdx: integer; const Fmt: RawUtf8;
       const Args: array of const);
@@ -633,7 +628,7 @@ type
 constructor TYamlToJson.Create;
 begin
   inherited Create;
-  fOut := TJsonWriter.CreateOwnedStream(fOutBuf);
+  fOut := TJsonWriter.CreateOwnedStream;
   fMaxDepth := YamlMaxDepth;
   if fMaxDepth < 4 then
     fMaxDepth := 4; // sanity floor: allow at least a few nested structures
@@ -903,7 +898,6 @@ function TYamlToJson.CollectBlockScalar(MinIndent: integer; Folded: boolean;
   Chomp: AnsiChar; ExplicitIndent: integer): RawUtf8;
 var
   tmp: TTextWriter;
-  wbuf: TTextWriterStackBuffer;
   i, blockIndent, startIdx: integer;
   raw, content: RawUtf8;
   lineContent: RawUtf8;
@@ -965,7 +959,7 @@ begin
     end;
     exit;
   end;
-  tmp := TTextWriter.CreateOwnedStream(wbuf);
+  tmp := TTextWriter.CreateOwnedStream;
   try
     blankRun := 0;
     prevWasContent := false;
@@ -1146,7 +1140,6 @@ var
 
 var
   tmp: TTextWriter;
-  wbuf: TTextWriterStackBuffer;
   cur: RawUtf8;
 begin
   if rest = '' then
@@ -1157,7 +1150,7 @@ begin
   if ScanQuoteClosed(rest, {skipOpeningQuote=}true) then
     exit; // single-line, already closed
   // multi-line merge
-  tmp := TTextWriter.CreateOwnedStream(wbuf);
+  tmp := TTextWriter.CreateOwnedStream;
   try
     cur := rest;
     while true do
@@ -1933,7 +1926,6 @@ type
   TVariantToYaml = class
   private
     fOut: TJsonWriter;
-    fOutBuf: TTextWriterStackBuffer;
     fOptions: TYamlWriterOptions;
     procedure WriteValue(const v: variant; Indent: integer);
     procedure WriteBlockMap(const dv: TDocVariantData; Indent: integer);
@@ -1953,7 +1945,7 @@ type
 constructor TVariantToYaml.Create(Options: TYamlWriterOptions);
 begin
   inherited Create;
-  fOut := TJsonWriter.CreateOwnedStream(fOutBuf);
+  fOut := TJsonWriter.CreateOwnedStream;
   fOptions := Options;
 end;
 
--- a/src/core/mormot.core.yaml.pas
+++ b/src/core/mormot.core.yaml.pas
@@ -221,16 +221,20 @@ end;
 
 function IsYamlNull(const S: RawUtf8): boolean;
 begin
-  result := (S = '') or (S = '~') or (S = 'null') or
-            (S = 'Null') or (S = 'NULL');
+  // YAML 1.2 core schema: null has three case-insensitive spellings,
+  // the '~' shortcut, or the empty string
+  result := (S = '') or
+            (S = '~') or
+            IdemPropNameU(S, 'null');
 end;
 
 function IsYamlBool(const S: RawUtf8; out V: boolean): boolean;
 begin
+  // YAML 1.2 core schema accepts any case for the boolean literal
   result := true;
-  if (S = 'true') or (S = 'True') or (S = 'TRUE') then
+  if IdemPropNameU(S, 'true') then
     V := true
-  else if (S = 'false') or (S = 'False') or (S = 'FALSE') then
+  else if IdemPropNameU(S, 'false') then
     V := false
   else
     result := false;
@@ -1880,8 +1884,8 @@ begin
   Doc.Clear;
   src := Yaml;
   // strip UTF-8 BOM regardless of source (file or in-memory) for consistency
-  if (length(src) >= 3) and (src[1] = #$EF) and
-     (src[2] = #$BB) and (src[3] = #$BF) then
+  if (length(src) >= 3) and
+     (PCardinal(pointer(src))^ and $00ffffff = BOM_UTF8) then
     delete(src, 1, 3);
   if Options = [] then
     opt := JSON_FAST_FLOAT + [dvoInternNames]
@@ -2012,8 +2016,8 @@ begin
   begin
     if IsYamlNull(S) then
       needsQuote := true
-    else if (S = 'true') or (S = 'True') or (S = 'TRUE') or
-            (S = 'false') or (S = 'False') or (S = 'FALSE') then
+    else if IdemPropNameU(S, 'true') or
+            IdemPropNameU(S, 'false') then
       needsQuote := true
     else if IsYamlFloat(S) then
       needsQuote := true
--- a/src/core/mormot.core.yaml.pas
+++ b/src/core/mormot.core.yaml.pas
@@ -80,6 +80,19 @@ function VariantToYaml(const Doc: variant;
 procedure SaveVariantToYamlFile(const Doc: variant; const FileName: TFileName;
   Options: TYamlWriterOptions = []);
 
+/// convert YAML 1.2 UTF-8 text directly into a JSON RawUtf8
+// - convenience wrapper around the internal parser's JSON buffer, useful for
+//   pipelines that feed further JSON-based processing (e.g. RTTI settings,
+//   JsonToObject, LoadJson) without going through TDocVariantData
+// - returns '' on parse failure (inspect EYamlException for details)
+function YamlToJson(const Yaml: RawUtf8): RawUtf8;
+
+/// convert JSON UTF-8 text into YAML 1.2 UTF-8 text
+// - pipes the JSON through TDocVariantData then VariantToYaml
+// - useful for converting existing JSON settings files or API payloads to YAML
+function JsonToYaml(const Json: RawUtf8;
+  Options: TYamlWriterOptions = []): RawUtf8;
+
 
 var
   /// maximum YAML nesting depth before the parser raises EYamlException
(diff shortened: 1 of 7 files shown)


src/core/mormot.core.yaml.pas
```diff
@@ -0,0 +1,2308 @@
...
+    Indent: integer; // count of leading space characters
+    Content: RawUtf8; // trimmed of leading spaces and trailing \r
+    Raw: RawUtf8; // original line with trailing \r stripped
+  end;
+  TYamlLines = array of TYamlLine;
+
+function CountLeadingSpaces(const S: RawUtf8): integer;
+var
+  p: PUtf8Char;
+  n: integer;
```
synopse, line 122: use PtrInt instead of integer for lengths or positions

src/core/mormot.core.yaml.pas
```diff
@@ -0,0 +1,2308 @@
...
+    if (p = stop) or (p^ = #10) then
+    begin
+      eol := p;
+      // strip trailing \r if present
+      if (eol > lineStart) and ((eol - 1)^ = #13) then
+        FastSetString(raw, lineStart, eol - 1 - lineStart)
+      else
+        FastSetString(raw, lineStart, eol - lineStart);
+      if idx >= length(Lines) then
+        SetLength(Lines, length(Lines) * 2);
```
synopse: use NextGrow() instead of *2

src/core/mormot.core.yaml.pas
```diff
@@ -0,0 +1,2308 @@
...
+      Lines[idx].Raw := raw;
```
synopse: use a local PYamlLines instead of Lines[idx].

src/core/mormot.core.yaml.pas
```diff
@@ -0,0 +1,2308 @@
...
+end;
+
+// split Yaml into lines with indent metadata
+// - accepts both LF and CRLF terminators; trailing \r is stripped
+procedure SplitYamlLines(const Yaml: RawUtf8; out Lines: TYamlLines);
+var
+  p, lineStart, eol: PUtf8Char;
+  stop: PUtf8Char;
+  idx, len: integer;
+  raw: RawUtf8;
```
synopse, line 137: use a local current: PYamlLines instead of Lines[idx].
and directly set current^.raw instead of a transient variable

src/core/mormot.core.yaml.pas
```diff
@@ -0,0 +1,2308 @@
...
+begin
+  Lines := nil;
+  SetLength(Lines, 16);
```
synopse: Lines := nil not needed by compiler design.
SetLength() not needed if we use NextGrow() below.

src/core/mormot.core.yaml.pas
```diff
@@ -0,0 +1,2308 @@
...
+    end;
+    if digitsExp = 0 then
+      exit;
+  end;
+  if p <> stop then
+    exit; // trailing junk
+  result := sawDot or sawExp; // must be float-shaped, not just int
+end;
+
+function ParseQuotedString(const Frag: RawUtf8; out V: RawUtf8): boolean;
```
synopse, line 419: do not reinvent the whell: use faster mormot.core.json functions

src/core/mormot.core.yaml.pas
```diff
@@ -0,0 +1,2308 @@
...
+  Options: TDocVariantOptions): boolean;
+var
+  conv: TYamlToJson;
+  json, src: RawUtf8;
+  opt: TDocVariantOptions;
+begin
+  result := false;
+  Doc.Clear;
+  src := Yaml;
+  // strip UTF-8 BOM regardless of source (file or in-memory) for consistency
```
synopse, line 1886: there is a function in mormot.core.unicode for this

synopse:
Very nice code.
The fact that there are good test coverage is very good.

Of course, it is not as fast and optimized as our mORMot JSON parsers, but it is not needed for YAML usage (settings or description).

I have added some (quick / one pass) review comments.

src/core/mormot.core.yaml.pas
```diff
@@ -0,0 +1,2308 @@
...
+
+
+var
+  /// maximum YAML nesting depth before the parser raises EYamlException
+  // - default 512 is ample for real-world OpenAPI specs; raise for
+  //   pathologically deep inputs
+  // - converts would-be EStackOverflow into a clean EYamlException with
+  //   line info; guards both block and flow recursive descent
+  YamlMaxDepth: integer = 512;
+
```
synopse, line 104: It would be very convenient to expose also YamlToJson() and JsonToYaml() functions, e.g. to be used with other part of the framework, like settings and RTTI.

src/core/mormot.core.yaml.pas
```diff
@@ -0,0 +1,2308 @@
...
+type
+  TYamlToJson = class
+  private
+    fLines: TYamlLines;
+    fCount: integer;
+    fIdx: integer;
+    fDepth: integer;
+    fMaxDepth: integer;
+    fOut: TJsonWriter;
+    fOutBuf: TTextWriterStackBuffer;
```
synopse: don't put it as member, just let TJsonWriter.CreateOwnedStream() allocates its own (bigger) buffer

src/core/mormot.core.yaml.pas
```diff
@@ -0,0 +1,2308 @@
...
+end;
+
+
+{ ----- TDocVariant -> YAML writer ----------------------------------------- }
+
+type
+  TVariantToYaml = class
+  private
+    fOut: TTextWriter;
+    fOutBuf: TTextWriterStackBuffer;
```
synopse: same as above: no buffer here

src/core/mormot.core.yaml.pas
```diff
@@ -0,0 +1,2308 @@
...
+  fOut.Free;
+  inherited Destroy;
+end;
+
+procedure TVariantToYaml.WriteIndent(N: integer);
+var
+  i: integer;
+begin
+  for i := 1 to N do
+    fOut.Add(' ');
```
synopse: fOut.AddSpace()

src/core/mormot.core.yaml.pas
```diff
@@ -0,0 +1,2308 @@
...
+        needsQuote := true;
+        break;
+      end;
+    end;
+  if not needsQuote then
+  begin
+    fOut.AddNoJsonEscape(pointer(S), n);
+    exit;
+  end;
+  // emit as double-quoted with JSON-like escapes
```
synopse: fOut.AddJsonEscape()

src/core/mormot.core.yaml.pas
```diff
@@ -0,0 +1,2308 @@
...
+end;
+
+procedure TVariantToYaml.WriteScalar(const v: variant);
+var
+  vt: word;
+  s: RawUtf8;
+begin
+  vt := TVarData(v).VType;
+  case vt of
+    varEmpty, varNull:
```
synopse, line 2089: most of the types are better handled directly with fOut.AddVariant()

src/core/mormot.core.yaml.pas
```diff
@@ -0,0 +1,2308 @@
...
+  end;
+end;
+
+procedure TVariantToYaml.WriteFlow(const dv: TDocVariantData);
+var
+  i: integer;
+begin
+  if dv.Kind = dvObject then
+  begin
+    fOut.Add('{');
```
synopse: AddDirect() here and where possible

src/core/mormot.core.yaml.pas
```diff
@@ -0,0 +1,2308 @@
...
+    for i := 0 to dv.Count - 1 do
+    begin
+      if i > 0 then
+        fOut.AddShort(', ');
```
synopse: no space needed after : ,

zen010101:
Thanks for the review — all 15 points handled, pushed as 4 refactor commits on top of the original `feat`. One caveat worth flagging up front: an initial sweep replacing every `Add('c')` with `AddDirect('c')` triggered an access violation on the 9.2 MB GitHub REST spec, because `AddDirect` / `AddComma` are documented "no buffer check, to be called within the 16-byte overhead after a regular `Add()`". That attempt was rolled back; only `WriteIndent` was switched, to the buffer-checked `AddChars(' ', N)` primitive. A more careful per-call-site pass (wrapping each `AddDirect` in a preceding buffer-checked write) is deferred as a follow-up.

**Commit map**

| Review | Commit | Note |
| --- | --- | --- |
| #1 `YamlToJson` / `JsonToYaml` top-level API | `718473c` | exposed for RTTI / settings pipelines |
| #2 `PtrInt` | `c082c45`, `f0b829e` | writer body + `SplitYamlLines` |
| #3, #6 local `PYamlLine` pointer | `c082c45` | applied in `SplitYamlLines` |
| #4 drop `Lines := nil` + pre-`SetLength` | `c082c45` | |
| #5 `NextGrow()` | `c082c45` | |
| #7 `mormot.core.json` helpers in `ParseQuotedString` | — | kept: YAML 1.2 §5.7 escapes (`\a \v \e \x \0 \<space>`) are a superset of JSON's; `GetJsonField` / `GetJsonString` would reject valid YAML input |
| #8, #10 no member buffer | `c082c45` | all 5 writer sites now use `CreateOwnedStream` with no argument |
| #9 `mormot.core.unicode` BOM helper | `fd627d1` | switched to `PCardinal(p)^ and $00ffffff = BOM_UTF8` |
| #11 `AddSpace()` | `f0b829e` | `WriteIndent` → `fOut.AddChars(' ', N)` (used `AddChars` since `AddSpace` / `AddSpaces` don't seem to exist) |
| #12 `AddJsonEscape()` | `718473c` | `WriteYamlString` delegates to `AddJsonString` (quotes + escape in one call) |
| #13 `AddVariant()` | `718473c` | numeric types via `fOut.AddVariant(v, twNone)`; string-like still routes through `WriteYamlString` for YAML-specific quoting |
| #14 `AddDirect()` where possible | `f0b829e` | sweep rolled back (see caveat); only the buffer-checked `AddChars` path applied |
| #15 no space after `:` `,` | — | `:` kept with whitespace because YAML 1.2 flow grammar requires it (`{k:v}` parses as a single plain scalar key, not a mapping). Space after `,` is optional — kept for now; happy to drop in a follow-up |

**Validation**

- `TTestCoreYaml`: 0 / 179
- `TNetworkProtocols.OpenAPI` (incl. `_OpenApiYamlDispatch`): 0 / 779
- Full `mormot2tests` regression: all pass (one pre-existing host-dependent `Dmi smbios` OEM check reproduces on a clean upstream checkout too, unrelated)
- Public OpenAPI stress re-run: Petstore v3, DigitalOcean, Stripe (6 MB → 27 K-line client), GitHub REST (9.2 MB → 62 K-line client) — all four byte-identical to the pre-refactor baseline (modulo the generation-date header line)

src/core/mormot.core.yaml.pas
zen010101: Done in 718473c — `YamlToJson` / `JsonToYaml` exposed alongside `YamlToVariant` / `VariantToYaml`.

src/core/mormot.core.yaml.pas
zen010101: Done in c082c45 and f0b829e — `SplitYamlLines` and the writer both switched to `PtrInt`.

src/core/mormot.core.yaml.pas
zen010101: Done in c082c45 — `SplitYamlLines` now uses a `current: ^TYamlLine` local to replace the repeated `Lines[idx].` access.

src/core/mormot.core.yaml.pas
zen010101: Done in c082c45.

src/core/mormot.core.yaml.pas
zen010101: Done in c082c45 — `NextGrow(idx)` replaces `length(Lines) * 2`.

src/core/mormot.core.yaml.pas
zen010101: Done in c082c45 — same local pointer as above.

src/core/mormot.core.yaml.pas
zen010101: Kept as-is: YAML 1.2 §5.7 double-quoted escapes (`\a \v \e \x \0 \<space>`) are a superset of JSON's set, so `GetJsonField` / `GetJsonString` would reject valid YAML input.

src/core/mormot.core.yaml.pas
zen010101: Done in c082c45 — `fOutBuf` member removed; all 5 writer sites use `CreateOwnedStream` with no argument.

src/core/mormot.core.yaml.pas
zen010101: Done in fd627d1 — switched to `PCardinal(pointer(src))^ and $00ffffff = BOM_UTF8`, the `mormot.core.unicode` idiom.

src/core/mormot.core.yaml.pas
zen010101: Done in c082c45 — same treatment as `TYamlToJson.fOutBuf`.

src/core/mormot.core.yaml.pas
zen010101: Done in f0b829e — `WriteIndent` collapses to `fOut.AddChars(' ', N)`. Used `AddChars` because `AddSpace` / `AddSpaces` don't appear to exist on the writer.

src/core/mormot.core.yaml.pas
zen010101: Done in 718473c — `WriteYamlString` delegates to `fOut.AddJsonString(S)` (quotes + escape in one call). The 30-line hand-rolled escape loop is gone.

src/core/mormot.core.yaml.pas
zen010101: Done in 718473c — numeric variants (byte / word / integer / int64 / single / double / currency) go through `fOut.AddVariant(v, twNone)`. String-like types still need YAML-specific quoting heuristics so they route through `WriteYamlString`.

src/core/mormot.core.yaml.pas
zen010101: Attempted but rolled back: a sweep converting hot-path `Add('c')` / `AddShort(', ')` to `AddDirect` / `AddComma` triggered an access violation on the 9.2 MB GitHub REST spec, since both are documented "no buffer check — within the 16-byte overhead after a regular `Add()`". Kept the buffer-checked `AddChars` path in `WriteIndent` (f0b829e). A selective per-call-site pass can be a follow-up.

src/core/mormot.core.yaml.pas
zen010101: `:` kept with trailing space in flow style because YAML 1.2 flow grammar requires whitespace after the key separator (`{k:v}` parses as a single plain scalar key, not a mapping). Space after `,` is optional — kept the existing `AddShort(', ')` for now; happy to drop it in a follow-up.

test/test.core.yaml.pas
```diff
@@ -0,0 +1,448 @@
...
+  CheckUtf8(ok, 'YamlToVariant returned false for %', [Name]);
+  actual := doc.ToJson;
+  CheckEqual(actual, ExpectedJson, FormatUtf8('golden "%"', [Name]));
+end;
+
+procedure TTestCoreYaml.ExpectRaise(const Name, Yaml: RawUtf8);
+var
+  doc: TDocVariantData;
+  raised: boolean;
+begin
```
synopse, line 255: To avoid exception logged into a local file, you may set `TSynLog.Family.ExceptionIgnoreCurrentThread := true` temporarly during the tests.

Here and everywhere the `expect raise := true` pattern is used.

There is a `TSynTestCase.CheckRaised()` method for this purpose.

synopse (changes requested):
Nice!

One question: where is the "Public OpenAPI stress re-run" done?

synopse:
Yes, `AddDirect()` could be nasty to use. Since we don't search for absolute performance, don't change it, I could do it manually with proper knowledge of safety.

zen010101:
The stress run was done locally, not from the test suite — the specs themselves (~15 MB total, 9.2 MB for GitHub REST alone) aren't really suitable for committing. Reproduction recipe:

```bash
mkdir /tmp/openapi-stress && cd /tmp/openapi-stress
curl -fsSL -o petstore3.yaml    https://petstore3.swagger.io/api/v3/openapi.yaml
curl -fsSL -o digitalocean.yaml https://raw.githubusercontent.com/digitalocean/openapi/main/specification/DigitalOcean-public.v2.yaml
curl -fsSL -o stripe.yaml       https://raw.githubusercontent.com/stripe/openapi/master/openapi/spec3.yaml
curl -fsSL -o github.yaml       https://raw.githubusercontent.com/github/rest-api-description/main/descriptions/api.github.com/api.github.com.yaml
mkdir -p out
for f in *.yaml; do
  mopenapi "$f" "${f%.yaml}" --concise "--DestinationFolder=out/${f%.yaml}/"
done
wc -l out/*/*.api.pas
```

Happy to land a CI-friendly variant (fetched with a cached / gated helper) in a follow-up PR if it'd be useful.

And thanks for the note on `AddDirect()` — understood, leaving the buffer-safety pass to you.

synopse:
Thanks!
Merging and cleaning.

synopse:
I guess the "mini" files added in test/data were expected to be included in the OpenAPI zip tests?
I will add them to https://synopse.info/files/openapi-ref.zip and include them and delete them from github.

synopse:
I have moved the unit into mormot.core.fmt.pas since there was room in it, and YAML would eventually be needed by TJsonFileSettings which is located in this unit.

zen010101:
Makes sense — and it lines up exactly with the integration direction you sketched in #467.

PR #472, opened 2026-05-22
core: fix Delphi Linux x64 SOA ABI for by-ref function results

AntoineGS:
This resolves an Access Violation when using SOA, which stems from a difference in the order of registers on FPC/Windows versus the LLVM compiler on Linux64.
Feel free to make changes.

### DIFF
--- a/src/core/mormot.core.interfaces.pas
+++ b/src/core/mormot.core.interfaces.pas
@@ -1842,6 +1842,18 @@ SYSV aarch64:
  https://c9x.me/compile/bib/abi-arm64.pdf
 }
 
+{$ifdef ISDELPHI}
+  {$ifdef ABISYSVX64}
+    // Delphi's LLVM Linux x64 compiler returns a by-ref result (managed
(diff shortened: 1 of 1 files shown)


synopse (approved):
Nice!

I will just rewrite the SelfFromInterface logic here but anyway it sounds fine.
Thanks!

synopse:
Can you try to run the regression tests? 

I guess we should add some other parameters to the test.

AntoineGS:
Running the tests uncovered some more AVs and possibly a deadlock in the MultiThreaded tests (with some minor compilation errors too).
I'll do some digging today and Monday, and should have more to share next week.

myonlylonely:
@AntoineGS does these more AVs only exists in SOA modules?

AntoineGS:
My PR with the fixes for what I found (and to get all tests to pass) was already merged to main/master earlier today. However there is a compile time error currently which will probably get resolved whenever @synopse gets a crack at it.

PR #473, opened 2026-05-25
Fixes to Delphi Linux64 Regressions

AntoineGS:
This is a followup on our test suite discussion on the previous PR.
Running the tests ended up showing quite a few issues specific to Delphi Linux, most having possible production impacts.
Each of of the commits in here (there are quite a few, I can create a PR for each one if you prefer, not problem) resolves failures in the test suite, with the final count being 0 failures.

I heavily leaned on Claude for the investigations and initial solutions, but I have reviewed them multiple times over the last two days to understand the changes and tweak some of the logic. However the crypto fix I am not not as confident, mostly as to what the impacts of changing this are in the overall framework; and the assembly changes, I have never worked in assembly so I took it as-is.
I added tests on the commits that could be tested deterministically, but some of them simply make existing tests pass or stop deadlocking.
There are a lot of comments and long commit messages, mostly because I wanted to understand the changes myself and kept adding to them but I also figured you could trim them down if things looked obvious to you.

Tests pass both on Windows and Delphi Linux after this (I have not tested other targets).
The first two resolve compilation issues, the rest are runtime issues.

Edit: The more I think about this the more I think I should split these up into different PRs so we can discuss the changes separately if needed, just say the word :)


synopse (approved):
Very nice!

synopse:
Very nice!
Especially the low-level asm fixes, which were far from obvious.

Some minor refactoring (no functional change):

https://github.com/synopse/mORMot2/commit/0bd091b5f
- replace UnixSocketFileDelete() with fpunlinka() on FPC/POSIX for simplicity.

https://github.com/synopse/mORMot2/commit/5b946a532
- move TCP_KEEPIDLE constants to mormot.core.os.delphi.pas

https://github.com/synopse/mORMot2/commit/a19204b08
- simplified code on targets with no exception interception

PR #475, opened 2026-06-03
What is the basic set of types, classes and configuration I should reach for by default?

joaopauloschuler:
# Recommended Patterns & Default Toolbox

*A Possible Pattern for Using mORMot 2*

mORMot 2 is a large framework with many ways to do the same thing. That flexibility is a strength, but newcomers consistently ask the same question: **"What is the basic set of types, classes and configuration I should reach for by default?"** The reference chapters describe *what exists*; this guide describes *what to pick* when you have no specific reason to do otherwise.

It is organised in two parts, which are deliberately different in tone:

- **Part A — The Default Toolbox** is **framework-universal**. These are recommendations that hold for essentially any mORMot 2 project. Treat them as firm defaults: prefer the mORMot type unless you have a concrete reason not to.
- **Part B — An Opinionated Application Blueprint** is **one proven way to structure a service**, not a framework rule. It is a coherent, batteries-included starting point. Adopt it wholesale for a new service, or cherry-pick what is applicable to your project.


docs/mORMot2-SAD-Recommended-Patterns.md
```diff
@@ -0,0 +1,223 @@
...
+RestServer.ServiceDefine(TUserService, [IUserQuery, IUserCommand], sicShared);
+
+// 4. Authentication: enable it explicitly (JWT / mORMot auth)
+RestServer.AuthenticationRegister(TRestServerAuthenticationDefault);
+
+// 5. HTTP: expose over HTTP/HTTPS
+HttpServer := TRestHttpServer.Create('8080', [RestServer], '+', HTTP_DEFAULT_MODE);
+
+// Logging: turn it on early (see Chapter 25)
+TSynLog.Family.Level := LOG_VERBOSE;
```
synopse: Since it is early, perhaps put at first position before Model.
The other steps already start logging content.

docs/mORMot2-SAD-Recommended-Patterns.md
```diff
@@ -0,0 +1,223 @@
...
+
+### A.6. Minimum recommended server configuration
+
+A sane default server wires five things: a model, a storage server bound as `IRestOrm`, your services, an HTTP server, and logging. A minimal shape:
+
+```pascal
+// 1. Model: list every TOrm class the server persists
+Model := TOrmModel.Create([TUser, TInvoice]);
+
+// 2. Storage: embedded SQLite is the default system of record
```
synopse: Defining both the working DB and the interface-based services on the same RestServer instance is easy but could lead to security issues since it exposes the DB Model via REST endpoints, and confusion.
Use the "persistence" service pattern (see B.1 below), perhaps behind a server-side `I*Persistence` interface, with its own TRestServerDB instance, and keep only interface-based services at the main RestServer + HttpServer layer.
The main RestServer could be just a TRestServerFullMemory with a void model.

docs/mORMot2-SAD-Recommended-Patterns.md
```diff
@@ -0,0 +1,223 @@
...
+- **Folder layout.** Keep the persistence boundary visible: lowercase `dom/`, `infra/`, `app/`, `tests/`, one subfolder per entity. Don't merge domain and infrastructure into one folder.
+- **Result handling.** A shared base enum (e.g. `TServiceResult`: success / invalid-request / not-found / denied / db-error) plus optional per-service extensions — stops the proliferation of near-identical result enums.
+- **Auth on the consumer side.** Extract a single guard helper so every service does one `Guard.Require(Token, …)` call instead of a repeated multi-step token ritual.
+- **Tests.** A `tests/` folder per service; mapper procedures are the cheapest, highest-value test target — exercise them first.
+
+### B.7. Default starting point for a new service
+
+1. `TOrm` per entity (in `infra/`), doubling as the domain object.
+2. A family of (possibily `packed record`) DTOs (in `app/<Entity>/`), registered with `Rtti.RegisterFromText` in the unit's `initialization`.
+3. Paired `IXxxCommand` + `IXxxQuery` interfaces descending from `IInvokable`.
```
synopse: CQRS should be only one possibility. Not a pattern to always follow.

docs/mORMot2-SAD-Recommended-Patterns.md
```diff
@@ -0,0 +1,223 @@
...
+4. Service implementations as `TInjectableObjectRest` subclasses, `IRestOrm` injected via published property.
+5. `sicShared` instance mode unless profiling says otherwise.
+6. Standard daemon: load settings, build the `TOrmModel`, create the persistence server, wire repositories, register services, start `TRestHttpServer`.
+7. A single auth guard consuming tokens from the shared auth service.
+8. `RawUtf8` everywhere data crosses a boundary; `string` only where VCL/LCL forces it.
+9. A `tests/` folder, starting with mapper round-trip tests.
+
```
synopse, line 586: Perhaps add something about layers and architecture.
Especially, a nice pattern is the Application layer definition. You don't expose directly all the available service entities and methods (`TOrm` and interfaces). You start from the actual client needs (each screen, each step of the client workflow), then you expose a set of per-application layers with DTO as reduced entities and some spot-on interface-based services.
The Application layer would orchestrate between one or several lower-level services on the server side (e.g. authentication, offers, orders, invoice, accounting, management, audit trail...) which may be Micro-Services, i.e. stand-alone SOLID service with its own database.

synopse (commented):
Very nice!
I made some comments to the document.

I guess the goal is to help any Human or AI follow the best patterns?

joaopauloschuler:
@synopse , thank you so much for your feedback! I'll update the md file and FUP.

joaopauloschuler:
BTW, you are exactly right: I do intend to have a blueprint (a set of defaults) for both myself and AI.

docs/mORMot2-SAD-Recommended-Patterns.md
joaopauloschuler: Proposal: https://github.com/joaopauloschuler/mORMot2/commit/aa3124ab8e11bdbdab4264d02abd424475b64d4b

docs/mORMot2-SAD-Recommended-Patterns.md
joaopauloschuler: Done: https://github.com/joaopauloschuler/mORMot2/commit/87bef3c0032efb6285b6b6ed4d48f6482af7b899

docs/mORMot2-SAD-Recommended-Patterns.md
joaopauloschuler: Before I commit anything, what do you think about the following please?

A sane default server is **two REST servers, not one**. The split is deliberate: it makes the system secure *by construction* rather than by configuration you have to remember to get right. Defining the working database and the interface-based services on a *single* `TRestServer` is easy, but it exposes your `TOrm` model as REST CRUD endpoints (`/root/TUser`, `/root/TInvoice`, …) right next to your services — a second, business-logic-bypassing door onto the same data, reachable by any sufficiently-privileged session. The two-server shape removes that door entirely:

- a **private system-of-record server** (`TRestServerDB`, SQLite) that owns every business table *and* the user records — and is **never handed to an HTTP server**, so it has no network surface at all;
- a **public edge server** (`TRestServerFullMemory`, void model) that terminates TLS, validates JWT, and hosts only your interface-based services.

```pascal
// === PRIVATE: the system of record. Owns the data. No network surface. ===
// 1. Model: every business TOrm + the auth tables live here, and ONLY here
DbModel := TOrmModel.Create([TUser, TInvoice, TAuthUser, TAuthGroup]);
DbServer := TRestServerDB.Create(DbModel, 'data.db');
DbServer.Server.CreateMissingTables;          // server-only call, at composition root
// DbServer is deliberately NOT passed to any TRestHttpServer below:
// it is reachable only in-process, as the injected private DB ORM (see B.5).

// === PUBLIC: the edge. Owns no data. Only services + JWT. ===
// 2. Void model: no TOrm classes => no ORM REST routes can exist at all
EdgeModel := TOrmModel.Create([]);
EdgeServer := TRestServerFullMemory.Create(EdgeModel);

// 3. Services: register interface-based services, injecting the private DB ORM
//    explicitly. The services run on EdgeServer, so their inherited Server.Orm
//    is the edge ORM (void, useless here) — bind the private DB ORM at the
//    composition root instead (the explicit form of B.5's DI). The shared-
//    instance ServiceDefine overload registers a sicShared instance you built.
EdgeServer.ServiceDefine(
  TUserService.Create(DbServer.Orm),          // inject the private DB ORM
  [IUserQuery, IUserCommand]);                // registered as sicShared
```

docs/mORMot2-SAD-Recommended-Patterns.md
joaopauloschuler: I will work on it.

docs/mORMot2-SAD-Recommended-Patterns.md
synopse: Perhaps investigate a little deeper:
* It is a concern only if the network layer is public on the Internet. For a local/corporate service, it is not such an issue.
* The "private" record is likely to publish not only the database model, but some services too, e.g. repository (e.g. as CQRS), authentication, main data model.
* The "public" service will use locally on the server the "private" database model and services, in the same process/instance, or could orchestrate, several service instances on the server network.
* If the "public" service is published over the Internet, it better reduces the scope of the endpoints, e.g. by focusing on what is needed for each application context. It is a common mistake on public services to publish too much: for instance, a logged end-user should be able to retrieve only its own data, not all the data.

synopse (changes requested):

docs/mORMot2-SAD-Recommended-Patterns.md
joaopauloschuler: @synopse , what you say makes sense ...

Before I commit, this is the updated version for you to consider please:

### A.6. Minimum recommended server configuration

A sane default server wires a handful of deliberate decisions: a model, a storage server bound as `IRestOrm`, your services, authentication, an HTTP server, and logging. **Every one of these is a named decision — none should be left implicit.**

One of those decisions is *topology*: how many REST servers exist and what is reachable from the network. Choose it by deployment context, not by habit:

- **A.6.1** — a single server, the right default for a **trusted / local** deployment.
- **A.6.2** — a private system-of-record plus a public edge, for an **Internet-facing or distributed** deployment.

**A cross-cutting rule that applies to both.** Topology reduces *exposure*; it never substitutes for *authorization*. Scope every read to the caller: a logged-in end user must retrieve only their own rows, never the whole table. This is a service-design concern (filter each query to the session's identity at the service boundary) — no choice of server count fixes it. See [B.6](#b6-cross-cutting-standards).

#### A.6.1. Trusted / local deployment

For a service that is only reachable on a trusted network (local box, corporate LAN, in-cluster), a **single server** is correct and simplest. The ORM REST routes it exposes (`/root/TUser`, …) are not an unguarded second door: they sit behind the *same* session authentication as your services, and behind `TOrmAccessRights`, which lets you make a table read-only, forbid CRUD per-table, or expose none at all.

```pascal
// 1. Logging: turn it on early (see Chapter 25)
TSynLog.Family.Level := LOG_VERBOSE;
TSynLog.Family.PerThreadLog := ptIdentifiedInOnFile;

// 2. Model: list every TOrm class the server persists
Model := TOrmModel.Create([TUser, TInvoice]);

// 3. Storage: embedded SQLite is the default system of record
RestServer := TRestServerDB.Create(Model, 'data.db');
RestServer.Server.CreateMissingTables;       // server-only call, at composition root

// 4. Services: register interface-based services (default sicShared)
RestServer.ServiceDefine(TUserService, [IUserQuery, IUserCommand], sicShared);

// 5. Authentication: enable it explicitly (JWT / mORMot auth)
RestServer.AuthenticationRegister(TRestServerAuthenticationDefault);

// 6. Access rights: tighten or disable the ORM REST routes you do not want
//    exposed (per-table read-only / no-CRUD); services stay reachable.
RestServer.OnlyJsonRequests := true;
// e.g. RestServer.SetAccessRights(...) to lock down or drop ORM endpoints

// 7. HTTP: expose over HTTP/HTTPS
HttpServer := TRestHttpServer.Create('8080', [RestServer], '+', HTTP_DEFAULT_MODE);
```

Adjust per project: swap `TRestServerDB` for an external SQL/NoSQL backend behind the same `IRestOrm` abstraction (see [A.5](#a5-bind-to-irestorm-and-the-_safe-rule) and [B.1](#b1-storage-behind-an-interface)); choose your authentication scheme in [Chapter 21](mORMot2-SAD-Chapter-21.md); tune logging in [Chapter 25](mORMot2-SAD-Chapter-25.md).

#### A.6.2. Internet-facing or distributed deployment

When the edge is **public on the Internet**, or when you are **orchestrating several service instances** across a server network, split the topology to minimise the public surface; the gain is reduced exposure and a clean place to terminate TLS/JWT.

- A **private system-of-record** (`TRestServerDB`, SQLite or external SQL) that owns the business tables and the auth records, and is **never handed to an HTTP server** — so it has no network surface and is reachable only in-process. It still publishes services (repository / CQRS, authentication, the main data model); "private" means *not network-exposed*, not *data-only*.
- A **public edge** (`TRestServerFullMemory`, void model) that terminates TLS, validates JWT, and exposes only the **context-scoped** services each application actually needs. The edge consumes the private services either in-process (same instance) or over the network (distributed).

The edge coding is an *addition* to A.6.1, not a different world:

```pascal
// === PRIVATE: the system of record. Owns the data. No network surface. ===
// Same wiring as A.6.1 steps 1-6, but the resulting server is NOT passed to
// any TRestHttpServer: it stays reachable only in-process.
DbModel  := TOrmModel.Create([TUser, TInvoice, TAuthUser, TAuthGroup]);
DbServer := TRestServerDB.Create(DbModel, 'data.db');
DbServer.Server.CreateMissingTables;
// (register the private services on DbServer here: repository/CQRS, auth, ...)

// === PUBLIC: the edge. Owns no data. Only context-scoped services + JWT. ===
// Void model => no TOrm classes => no ORM REST routes can exist at all.
EdgeModel  := TOrmModel.Create([]);
EdgeServer := TRestServerFullMemory.Create(EdgeModel);

// Inject the private DB ORM explicitly at the composition root: the service
// runs on EdgeServer, whose inherited Server.Orm is the void edge ORM, so bind
// the private ORM here rather than relying on Server.Orm (the explicit form of
// B.5's DI). Publish only the services this application context needs.
EdgeServer.ServiceDefine(
  TUserService.Create(DbServer.Orm),          // inject the private DB ORM
  [IUserQuery, IUserCommand]);                // registered as sicShared

// Only the edge is handed to HTTP — the private server stays off the network.
HttpServer := TRestHttpServer.Create('443', [EdgeServer], '+', HTTP_DEFAULT_MODE);
```

Publishing too much is the classic Internet-facing mistake; the cross-cutting rule from A.6 — scope every read to the caller — is what actually prevents data leaks here. See [B.6](#b6-cross-cutting-standards).

joaopauloschuler: @synopse , what do you think?

synopse: It seems somewhat practical, but to my understanding it lacks a lot of software architecture principles, and re-invent some concepts like "public edge" and such. They are better named after the regular definitions e.g. from "Clean Architecture". It would also help understanding the idea behind, because both AI and humans can refer to the existing literature.

I see see two main designs possible: 
1) monolithic for simple projects made e.g. by one individual for a small scope.
2) orchestrated with an application layer and several sub-services.

Worth noting the lack of distinction between logical and physical services. mORMot can host the very same services remotely or locally, it is just a matter of instances initialization. So you can have physically a monolith server (a single exe) but logically an internal set of layered services, e.g. following KDD/DDD principles, CQRS, Event-Driven or MicroServices.

docs/mORMot2-SAD-Recommended-Patterns.md
joaopauloschuler: @synopse , should this comment be resolved?

docs/mORMot2-SAD-Recommended-Patterns.md
joaopauloschuler: @synopse , should this comment be resolved?

docs/mORMot2-SAD-Recommended-Patterns.md
```diff
@@ -0,0 +1,504 @@
...
+  // Remote: from a client pointed at the other instance.
+  if Rest.Services.Resolve(IUserQuery, svc) then
+    // From here the call site is identical, wherever svc actually runs:
+    profile := svc.GetProfile(userID);
+```
+
+Switching a service from local to remote is a matter of *instance initialization* — register it on a different server, point a client at a URI — **not a code change**. The layering you design in Part B therefore survives any topology you later deploy it on.
+
+**Make the switch a single, localized decision (the client-abstraction pattern).** Concentrate the local-vs-remote choice in one *client-abstraction* unit that every consumer depends on, with two interchangeable implementations. Consumer code never changes and never learns which one it got — the only thing that varies is which implementation unit gets wired up at startup:
+
```
synopse, line 159: Using pure abstract classes is IMHO overkill for such a presentation document.
The pattern to use is not such classes, but the easier and more classic "resolver" pattern, i.e. the mORMot `TInterfaceResolver.Resolve()` method. There is also a `TInjectableObject` and so on in mormot.core.interfaces.

docs/mORMot2-SAD-Recommended-Patterns.md
```diff
@@ -0,0 +1,504 @@
...
+// 4. Services: register interface-based services (default sicShared)
+RestServer.ServiceDefine(TUserService, [IUserQuery, IUserCommand], sicShared);
+
+// 5. Authentication: enable it explicitly (JWT / mORMot auth)
+RestServer.AuthenticationRegister(TRestServerAuthenticationDefault);
+
+// 6. Access rights: tighten or disable the ORM REST routes you do not want
+//    exposed (per-table read-only / no-CRUD); services stay reachable.
+RestServer.OnlyJsonRequests := true;
```
synopse: Hallucination I guess.

synopse (changes requested):
It is a huge task for sure...

docs/mORMot2-SAD-Recommended-Patterns.md
joaopauloschuler: I must have pressed CTRL+C + CTRL+V in the wrong place...

docs/mORMot2-SAD-Recommended-Patterns.md
joaopauloschuler: Agreed.

joaopauloschuler: @synopse , should this comment be resolved? I believe that I have already solved it.

docs/mORMot2-SAD-Recommended-Patterns.md
```diff
@@ -0,0 +1,674 @@
...
+```
+
+</details>
+
+<details>
+<summary><b>AppUserClientRemote.pas</b> — an HTTP client to the standalone server (the A.6.2 distributed layout)</summary>
+
+```pascal
+unit AppUserClientRemote;
+
```
synopse, line 236: add minimal `interface` with `UserResolver` global variable perhaps.

docs/mORMot2-SAD-Recommended-Patterns.md
```diff
@@ -0,0 +1,674 @@
...
+From there, three call shapes cover everything:
+
+```pascal
+UserMap.ToB(user, @dto);   // fill an existing DTO   from a TOrm   (ORM -> DTO)
+UserMap.ToA(user, @dto);   // fill an existing TOrm  from a DTO    (DTO -> ORM)
+dtoPtr := UserMap.ToB(user);          // allocate + fill a NEW DTO from a TOrm
+newUser := UserMap.ToA(@dto);         // allocate + fill a NEW TOrm from a DTO
+UserMap.ToArrayB(users, dtos);        // map a whole dynamic array at once
+```
+
```
synopse, line 447: Take into account property casing.
Most public APIs expect the JSON to have either a lower-cased first letter, or some other naming 
Consider `TRttiCustomProps.NameChangeCase()` to normalize the naming of the DTO record properties when serialized to JSON, e.g. with `scCamelCase`, `scSnakeCase` or  `scKebabCase`, depending on the common used convention in the existing system. Most languages are case-sensitive for their identifier, so it could make sense to make it explicit.

synopse (changes requested):

docs/mORMot2-SAD-Recommended-Patterns.md
joaopauloschuler: Thank you so much for your help!
This is an attempt to fix it: https://github.com/joaopauloschuler/mORMot2/commit/f1fba1c61fff3df3b758f9e54d26e170bd882b61

docs/mORMot2-SAD-Recommended-Patterns.md
joaopauloschuler: What do you think?
https://github.com/joaopauloschuler/mORMot2/commit/06e1b427cb84140719471c7adfae0f221c94e576

joaopauloschuler:
@synopse, only if you like the idea and if I find free time, would be in your interest me to code an example that follows these recommendations? Besides the current md file, we would have a working source code that follows these recommendations. I would make it work with lazbuild inside GitHub spaces so users would be able to test it without any local environment.

synopse:
> @synopse, we would have a working source code that follows these recommendations. 

That's a good idea, of course!

This was the idea of https://github.com/synopse/mORMot2/tree/master/ex/tdd-service AFAIR.
Perhaps to be finalized.

PR #477, opened 2026-06-10
mormot.core.os.pas SERVICE_ALL_ACCESS constant invalid 

GoranDespalatovic:
According to 
https://learn.microsoft.com/en-us/windows/win32/services/service-security-and-access-rights

SERVICE_ALL_ACCESS (0xF01FF)	Includes STANDARD_RIGHTS_REQUIRED in addition to all access rights in this table.

With old value (0x01FF) DeleteService returns ERROR_ACCESS_DENIED


synopse (approved):

synopse:
Thanks for the feedback!

PR #478, opened 2026-06-11
fix after adddynarr

dkounal:


synopse:
I guess it has been fixed by
https://github.com/synopse/mORMot2/commit/aad4c731af

Thanks for the feedback.
But I am not sure it is the root cause of the issue.
As I wrote in the forum, a dynamic array should never have length=0 with self<>nil. So there should be something wrong elsewhere.

PR #488, opened 2026-06-29
Fix handling Blob and DateTime type. Fix for small memory leak of…

TTomas:
Fix handling Blob and DateTime type. Fix for small memory leak of IBX cached interface (SetRetainInterfaces)


synopse (commented):
Small typos, but I will fix them.

Thanks for the feedback!

PR #497, opened 2026-07-21
net.client: add STARTTLS support to SendEmail

tobfel:
## Summary

Adds **STARTTLS** support to `SendEmail()` in `mormot.net.client`, implemented natively on top of `TCrtSocket.DoTlsAfter` (SChannel on Windows, OpenSSL on POSIX) - no external process.

Motivation: `SendEmail()` only supported plain SMTP (port 25) and *implicit* TLS, and the `TSmtpConnection` overload forced *implicit* TLS on port **587** - the STARTTLS submission port that most providers (Gmail, Office365, GMX...) expect and where implicit TLS is rejected.

This PR was reworked to a **single, focused commit** following the review feedback.

## Changes

- new `TSmtpTls = (stlsNone, stlsImplicit, stlsStartTls)` enum, replacing the former `TLS: boolean` parameter of both overloads;
- native STARTTLS via the `STARTTLS` command + `TCrtSocket.DoTlsAfter(cstaConnect)`;
- `TSmtpConnection` overload derives the mode from the port: 465 -> implicit, 587 -> STARTTLS;
- **replaced the `TLSIgnoreCertError` boolean by an optional `aTLS: PNetTlsContext`**, so the caller can specify any TLS option (as suggested in review);
- `EHLO` captured once; `AUTH PLAIN` used as a fallback when `AUTH LOGIN` is not advertised;
- **merged `Expect`/`Exec`/capability helpers into a single `Exec()` sub-function**;
- example uses `{$ifdef OSPOSIX}` consistently; ANSI encoding preserved (no stray non-ASCII byte).

Adds `ex/mail-starttls/mailtest.dpr` (+ `.lpi`). Builds with Delphi (dcc64) and FPC/Lazarus, and was verified to **cross-compile and link for x86_64-linux** as well. Verified end-to-end against a live provider on port 587.

## Note on certificate revocation (out of scope here, for discussion)

While testing against a real server I hit a Windows/SChannel-only issue worth flagging separately: some certificate chains carry **no reachable revocation info** (no CRL Distribution Point / OCSP), so SChannel aborts the handshake with `CRYPT_E_NO_REVOCATION_CHECK` even for otherwise-trusted certificates. The existing `SCH_CRED_IGNORE_REVOCATION_OFFLINE` only covers the "responder offline" case, not "no revocation info at all".

The only current workaround is `IgnoreCertificateErrors`, which disables *all* validation (trust chain, host name, expiration) - heavier than needed.

**Possible clean solution (happy to open as a separate PR if you're interested):** an opt-in `TNetTlsContext.IgnoreCertificateRevocation` flag that, on the SChannel client, additionally sets `SCH_CRED_IGNORE_NO_REVOCATION_CHECK` - turning revocation into best-effort (soft-fail) while still fully validating the trust chain, host name and expiration (the behaviour browsers use by default). No effect on OpenSSL, which does not check revocation by default. With the new `aTLS: PNetTlsContext` parameter above, callers could then request exactly that per email, without any Mail-specific flag.

I kept this out of the present PR to stay focused; let me know if you'd like it as a follow-up.

### DIFF
--- a/src/net/mormot.net.client.pas
+++ b/src/net/mormot.net.client.pas
@@ -2048,6 +2048,16 @@ type
   /// exception class raised by SendEmail() on raw SMTP process
   ESendEmail = class(ESynException);
 
+  /// how TLS should be applied when sending an email via SendEmail()
+  // - stlsNone: plain text connection (typically port 25)
+  // - stlsImplicit: TLS from the first byte of the connection (typically port 465)
+  // - stlsStartTls: plain connection upgraded to TLS via the STARTTLS command
+  // after EHLO (typically the submission port 587, as expected by most providers)
+  TSmtpTls = (
+    stlsNone,
+    stlsImplicit,
+    stlsStartTls);
+
 /// send an email using the SMTP protocol
 // - retry true on success
 // - the Subject is expected to be in plain 7-bit ASCII, so you could use
@@ -2056,11 +2066,15 @@ type
 // - you can optionally set another encoding charset or force TextCharSet='' to
 // expect the 'Content-Type:' to be set in Headers and Text to be the raw body
 // (e.g. a multi-part encoded message)
+// - set Tls to stlsImplicit for TLS-from-start (port 465) or stlsStartTls to
+// upgrade a plain connection via the STARTTLS command (port 587)
+// - optional aTLS could supply a TNetTlsContext to customize the TLS handshake
+// (e.g. aTLS^.IgnoreCertificateErrors), used for stlsImplicit and stlsStartTls
 function SendEmail(const Server, From, CsvDest, Subject: RawUtf8;
   const Text: RawByteString; const Headers: RawUtf8 = ''; const User: RawUtf8 = '';
   const Pass: RawUtf8 = ''; const Port: RawUtf8 = '25';
-  const TextCharSet: RawUtf8  =  'ISO-8859-1'; TLS: boolean = false;
-  TLSIgnoreCertError: boolean = false): boolean; overload;
+  const TextCharSet: RawUtf8  =  'ISO-8859-1'; Tls: TSmtpTls = stlsNone;
+  aTLS: PNetTlsContext = nil): boolean; overload;
 
 /// send an email using the SMTP protocol via a TSmtpConnection definition
 // - retry true on success
@@ -2070,11 +2084,12 @@ function SendEmail(const Server, From, CsvDest, Subject: RawUtf8;
 // - you can optionally set another encoding charset or force TextCharSet='' to
 // expect the 'Content-Type:' to be set in Headers and Text to be the raw body
 // (e.g. a multi-part encoded message)
-// - TLS will be forced if the port is either 465 or 587
+// - if Tls is left to stlsNone, it will be guessed from the port number:
+// stlsImplicit for port 465, or stlsStartTls for port 587
 function SendEmail(const Server: TSmtpConnection;
   const From, CsvDest, Subject: RawUtf8; const Text: RawByteString;
   const Headers: RawUtf8 = ''; const TextCharSet: RawUtf8  = 'ISO-8859-1';
-  TLS: boolean = false; TLSIgnoreCertError: boolean = false): boolean; overload;
+  Tls: TSmtpTls = stlsNone; aTLS: PNetTlsContext = nil): boolean; overload;
 
 /// convert a supplied subject text into an Unicode encoding
 // - will convert the text into UTF-8 and append '=?UTF-8?B?'
@@ -6107,29 +6122,43 @@ end;
 
 function SendEmail(const Server: TSmtpConnection;
   const From, CsvDest, Subject: RawUtf8; const Text: RawByteString;
-  const Headers, TextCharSet: RawUtf8; TLS, TLSIgnoreCertError: boolean): boolean;
+  const Headers, TextCharSet: RawUtf8; Tls: TSmtpTls; aTLS: PNetTlsContext): boolean;
 begin
+  if Tls = stlsNone then // guess the TLS mode from the well-known port numbers
+    if Server.Port = '465' then
+      Tls := stlsImplicit
+    else if Server.Port = '587' then
+      Tls := stlsStartTls;
   result := SendEmail(
     Server.Host, From, CsvDest, Subject, Text, Headers,
-    Server.User, Server.Pass, Server.Port, TextCharSet,
-    TLS or (Server.Port = '465') or (Server.Port = '587'), TLSIgnoreCertError);
+    Server.User, Server.Pass, Server.Port, TextCharSet, Tls, aTLS);
 end;
 
 function SendEmail(const Server, From, CsvDest, Subject: RawUtf8;
   const Text: RawByteString; const Headers, User, Pass, Port, TextCharSet: RawUtf8;
-  TLS, TLSIgnoreCertError: boolean): boolean;
+  Tls: TSmtpTls; aTLS: PNetTlsContext): boolean;
 var
   sock: TCrtSocket;
 
-  procedure Expect(const answer: RawUtf8);
+  // send the optional Command, read the whole (multi-line) answer, check the
+  // reply code, raise ESendEmail on mismatch, and return the full answer text
+  // - Command = '' is used to read the initial greeting or a previous send
+  function Exec(const Command, Answer: RawUtf8): RawUtf8;
   var
     res: RawUtf8;
   begin
+    if Command <> '' then
+    begin
+      sock.SockSend(Command);
+      sock.SockSendFlush;
+    end;
+    result := '';
     repeat
       sock.SockRecvLn(res);
+      result := result + res + #13#10;
     until (Length(res) < 4) or
-          (res[4] <> '-'); // - indicates there are other headers following
-    if IdemPChar(pointer(res), pointer(answer)) then
+          (res[4] <> '-'); // '-' indicates there are other lines following
+    if IdemPChar(pointer(res), pointer(Answer)) then
       exit;
     if res = '' then
       res := 'Undefined Error';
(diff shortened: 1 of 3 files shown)


synopse (changes requested):
This does not make any sense at all.
Both functions do NOT share the same  shortstirng buffer.

synopse (changes requested):
Why set SCH_CRED_REVOCATION_CHECK_CHAIN then set SCH_CRED_IGNORE_NO_REVOCATION_CHECK ?
Seems pretty weird to me.

ex/mail-starttls/mailtest.dpr
```diff
@@ -0,0 +1,138 @@
...
+{$I ..\..\src\mormot.defines.inc}
+
+{$ifdef OSWINDOWS}
+  {$apptype console}
+  {$R ..\..\src\mormot.win.default.manifest.res}
+{$endif OSWINDOWS}
+
+uses
+  {$I ..\..\src\mormot.uses.inc}
+  {$ifdef UNIX}
```
synopse: OSPOSIX ?

src/net/mormot.net.client.pas
```diff
@@ -6144,29 +6170,72 @@ function SendEmail(const Server, From, CsvDest, Subject: RawUtf8;
...
   P: PUtf8Char;
-  rec, ToList, head: RawUtf8;
+  rec, ToList, head, caps: RawUtf8;
 begin
   result := false;
   P := pointer(CsvDest);
   if P = nil then
     exit;
-  sock := SocketOpen(Server, Port, TLS, nil, nil, TLSIgnoreCertError);
+  // prepare a per-connection TLS context (so the flags stay local to this email)
```
synopse: the proper way may be to not use two booleans, but just the PNetTlsContext at caller level. Then you could specify whatever you need.

src/net/mormot.net.client.pas
```diff
@@ -6144,29 +6170,72 @@ function SendEmail(const Server, From, CsvDest, Subject: RawUtf8;
     Expect(answer)
   end;
 
+  // like Exec() but returns the whole (multi-line) server answer, e.g. for the
+  // EHLO reply so we can detect the advertised AUTH mechanisms
+  function ExecCap(const Command, answer: RawUtf8): RawUtf8;
+  var
```
synopse: some duplicate logic here for sure.
a single sub function could do both process

synopse (changes requested):

synopse:
Too many things at the same time, Claude confusion in the details, almost everything to rewrite by hand if I accept this pull request as such.

tobfel:
> This does not make any sense at all. Both functions do NOT share the same shortstirng buffer.

Before the SChannel - error message fix, i got, weird chars in the error. After the change the error was clear:

Before:
```
20260721 21161730  ! EXC   ESChannel {Message:"<mail.testserver.de>: HandshakeStep returned te at the domain controllerION_CHECKÜêƒû?x8Gpû?u0û?      ?$ý, System Error 1397 [Mutual Authentication failed. The server's password is out of date at the domain controller]"} [Main] at 472b5c mormot.net.sock.windows.inc TSChannelNetTls.ESChannelRaiseLastError (2049)   {16 3.82 4.83 2 21.4GB/30.7GB 66581b01}
20260721 21161730  ! EXC   ESChannel {Message:"<mail.testserver.de>: HandshakeStep returned te at the domain controllerION_CHECKÜêƒû?x8Gpû?u0û?      ?$ý, System Error 1397 [Mutual Authentication failed. The server's password is out of date at the domain controller]"} [Main] at 472b5c mormot.net.sock.windows.inc TSChannelNetTls.ESChannelRaiseLastError (2049)   {16 3.82 4.83 2 21.4GB/30.7GB 66581b01}
20260721 21161730  ! EXC   ENetSock {LastError:"nrUnknownError",Message:"TCrtSocket.DoTlsAfter: TLS failed [ESChannel <mail.testserver.de>: HandshakeStep returned te at the domain controllerION_CHECKÜêƒû?x8Gpû?u0û?       ?$ý, System Error 1397 [Mutual Authentication failed. The server's password is out of date at the domain controller]]"} [Main] at 476c26 mormot.net.sock.pas TCrtSocket.DoTlsAfter (6690)   {16 3.82 4.83 2 21.4GB/30.7GB 66581b01}
Error [ENetSock]: TCrtSocket.DoTlsAfter: TLS failed [ESChannel <mail.testserver.de>: HandshakeStep returned te at the domain controllerION_CHECKÜêƒû?x8Gpû?u0û?      ?$ý, System Error 1397 [Mutual Authentication failed. The server's password is out of date at the domain controller]]

```
AFTER:
```
20260721 21494924  ! EXC   ESChannel {Message:"<mail.testserver.de>: HandshakeStep returned 80092012 CRYPT_E_NO_REVOCATION_CHECK, System Error 1397 [Mutual Authentication failed. The server's password is out of date at the domain controller]"} [Main] at 472bac TSChannelNetTls.ESChannelRaiseLastError   {16 3.82 4.89 2 21.6GB/30.7GB 66581b01}
20260721 21494924  ! EXC   ESChannel {Message:"<mail.testserver.de>: HandshakeStep returned 80092012 CRYPT_E_NO_REVOCATION_CHECK, System Error 1397 [Mutual Authentication failed. The server's password is out of date at the domain controller]"} [Main] at 472bac TSChannelNetTls.ESChannelRaiseLastError   {16 3.82 4.89 2 21.6GB/30.7GB 66581b01}
20260721 21494924  ! EXC   ENetSock {LastError:"nrUnknownError",Message:"TCrtSocket.DoTlsAfter: TLS failed [ESChannel <mail.testserver.de>: HandshakeStep returned 80092012 CRYPT_E_NO_REVOCATION_CHECK, System Error 1397 [Mutual Authentication failed. The server's password is out of date at the domain controller]]"} [Main] at 476c86 TCrtSocket.DoTlsAfter   {16 3.82 4.89 2 21.6GB/30.7GB 66581b01}
Error [ENetSock]: TCrtSocket.DoTlsAfter: TLS failed [ESChannel <mail.testserver.de>: HandshakeStep returned 80092012 CRYPT_E_NO_REVOCATION_CHECK, System Error 1397 [Mutual Authentication failed. The server's password is out of date at the domain controller]]
```

src/net/mormot.net.sock.windows.inc
```diff
@@ -21,7 +21,7 @@ const
 
 const
   // default on Windows is 64, but we grow it to 512 for TPollSocketSelect
-  // as stated by Raymond Chen: "you can make FD_SET�SIZE bigger if you need to"
+  // as stated by Raymond Chen: "you can make FD_SET�SIZE bigger if you need to"
```
synopse: Invalid char here.

synopse (changes requested):

tobfel:
> Too many things at the same time, Claude confusion in the details, almost everything to rewrite by hand if I accept this pull request as such.

No Problem, only a try to add a feature. Because we need mail with starttls (tls already used with mormot) and i had problems with D13 / x64 i give claude code a try ;-) 

Result is now working for me, and i can say goodbye to Indy / TIdSMTP and the openssl-libs.

But sill some testing needed. And of course you write cleaner code than claude.

src/net/mormot.net.sock.windows.inc
tobfel: Yes, difficult to tell claude code to produce ANSI pas files. Not so nice with our german umlaute and Delphi7 :-(

tobfel:
Reworked into a single focused commit, addressing the review:

- **Two booleans -> `PNetTlsContext`**: `TLSIgnoreCertError` is gone; both overloads now take an optional `aTLS: PNetTlsContext`, so the caller can specify whatever TLS options they need.
- **Duplicate logic**: `Expect`/`Exec`/the capability reader are merged into a single `Exec()` sub-function (send optional command, read the multi-line reply, check the code, return the text).
- **`OSPOSIX`**: the example now uses `{$ifdef OSPOSIX}` consistently (cwstring guarded for FPC only).
- **Invalid char**: that was a UTF-8 byte my editor introduced; the file is back to ANSI and byte-identical to master outside the intended lines.
- The SChannel error-message change and the revocation flag are **dropped** from this PR to keep it focused. I left a note in the description about the `CRYPT_E_NO_REVOCATION_CHECK` case and a possible `TNetTlsContext.IgnoreCertificateRevocation` follow-up, if you're interested.

Re the shortstring point: you're right that it's compiler-specific; I only observed the corrupted output on Delphi 13 / Win64 and can't claim the mechanism in general, so I removed it here.

synopse (changes requested):
Can you please move the sample to https://github.com/synopse/mORMot2/tree/master/ex/ThirdPartyDemos/tobfel ?

Otherwise it is great now!
Thanks!

synopse:
> Re the shortstring point: you're right that it's compiler-specific; I only observed the corrupted output on Delphi 13 / Win64 and can't claim the mechanism in general, so I removed it here.

I think there is a buffer overflow in mormot.core.os.pas code.
Perhaps it is better with https://github.com/synopse/mORMot2/commit/b038312465711eadbd6780d6fb78a856fb1bae01
And it is clearly a compiler specific issue - perhaps about inlining.

synopse:
> And it is clearly a compiler specific issue - perhaps about inlining.

Which exact version of Delphi are you using? I remember 13.1 had issues with inlining so I added a VER370inlinebug conditional.

I have tried this to circumvent the issue:
https://github.com/synopse/mORMot2/commit/3b4ba7713

tobfel:
> > And it is clearly a compiler specific issue - perhaps about inlining.
> 
> Which exact version of Delphi are you using? I remember 13.1 had issues with inlining so I added a VER370inlinebug conditional.

Yes, D13.1 - i will try your changes and report back. Thanks.

tobfel:
> Can you please move the sample to https://github.com/synopse/mORMot2/tree/master/ex/ThirdPartyDemos/tobfel ?
> 
> Otherwise it is great now! Thanks!

I will do, thanks for the review. 

One Question, i had an long outstanding change for SendEmail:

I construct the Mail with Indy. TIdMessage / TIdMessageBuilderHtml and give the complete Mail-Raw Header-Data via idMailMessage.Headers.Text to your Send function. 
Problem with your Send: It is fixed filling "From:", "Subject: "

My solution was to check if the Header-Data do Start with "From:" or "Subject:" an if so i know that my Header-Data include this Infos. So i can construct a Mail with the Indy-Message object and use CC/BCC/ReplyTo without Cleanup the Header before "SendEmail". Should i try to do another PR for this? (no Indy-Refs of course)
 
 Another point: I commented out the last command (QUIT), because on TLS it results in Exceptions. Don't know if it is maybe server dependent. But i think the "." maybe closes the connection/session. If we use it it should maybe catch a Exception.

 Exec('.', '25');
 //Exec('QUIT', '22');

tobfel:
> > > And it is clearly a compiler specific issue - perhaps about inlining.
> > 
> > 
> > Which exact version of Delphi are you using? I remember 13.1 had issues with inlining so I added a VER370inlinebug conditional.
> 
> Yes, D13.1 - i will try your changes and report back. Thanks.

With you last changes i get some type errors on compilation: 

on D7:
[Fehler] mormot.core.os.windows.inc(718): Es gibt keine überladene Version von 'WinErrorShort', die man mit diesen Argumenten aufrufen kann
[Fehler] mormot.core.os.pas(6720): Es gibt keine überladene Version von 'OsErrorShort', die man mit diesen Argumenten aufrufen kann
[Fehler] mormot.core.os.pas(6927): Es gibt keine überladene Version von 'WinErrorShort', die man mit diesen Argumenten aufrufen kann
[Fehler] mormot.core.os.pas(7049): Es gibt keine überladene Version von 'OsErrorShort', die man mit diesen Argumenten aufrufen kann
[Fehler] mormot.core.os.pas(7057): Es gibt keine überladene Version von 'OsErrorShort', die man mit diesen Argumenten aufrufen kann

on D13:
[dcc64 Fehler] mormot.core.os.pas(851): E2439 Inline-Funktion darf kein Argument für ein Open Array haben

synopse:
This SendEmail() function was in fact just a small function for a specific notification purpose (send an email address to IT people on server instability).
Now it is used as a real Email sender logic kernel... and it is clearly under designed.

Perhaps it could be worth creating a dedicated TSmtpSend class with an associated TSmtpMessage class?
And I guess that instead of Indy code, we could reuse some part THttpMultiPartStream logic which is nice because it could stream some file content. A new TSmtpMultiPartStream = class(TNestedStreamReader) class could be defined.

But huge modification, not part of this pull request for sure.
Perhaps even a new dedicated unit. ;)

For ending the SMTP, i am not sure.
Sounds like QUIT is part of the protocol and should be correct.
https://chatgpt.com/s/t_6a6078facc988191bf20b32f14a5cb5b

synopse:
> on D13:
> [dcc64 Fehler] mormot.core.os.pas(851): E2439 Inline-Funktion darf kein Argument für ein Open Array haben

This is clearly a Delphi compiler limitation.
Should be fixed by https://github.com/synopse/mORMot2/commit/57be441d2

tobfel:
> > Can you please move the sample to https://github.com/synopse/mORMot2/tree/master/ex/ThirdPartyDemos/tobfel ?
> > Otherwise it is great now! Thanks!
> 
Done.

> My solution was to check if the Header-Data do Start with "From:" or "Subject:" an if so i know that my Header-Data include this Infos. So i can construct a Mail with the Indy-Message object and use CC/BCC/ReplyTo without Cleanup the Header before "SendEmail". Should i try to do another PR for this? (no Indy-Refs of course)

Created another PR (optimized) for this. Also added the option to give the mailtest.exe individual headers (-h "headername1: headervalue1" .. -h "headernameN: headervalueN" - so you can send/set also  Cc, Reply-To
> 
> Another point: I commented out the last command (QUIT), because on TLS it results in Exceptions. Don't know if it is maybe server dependent. But i think the "." maybe closes the connection/session. If we use it it should maybe catch a Exception.
> 
> Exec('.', '25'); //Exec('QUIT', '22');

encapsulated it in a try except, because it definitiv produce a exception on may smtp-test-servers.

tobfel:
> > on D13:
> > [dcc64 Fehler] mormot.core.os.pas(851): E2439 Inline-Funktion darf kein Argument für ein Open Array haben
> 
> This is clearly a Delphi compiler limitation. Should be fixed by [57be441d2](https://github.com/synopse/mORMot2/commit/57be441d2)

Yes D13 compiles again. D7 Problem is with the default-param overload. Shortest fix i think is to remove on default-param an change some calls of the function. I try to push a fix in my repo, after i finished my current task, if you want to have a look, if not fixed by you in the meanwhile.

tobfel:
> > > on D13:
> > > [dcc64 Fehler] mormot.core.os.pas(851): E2439 Inline-Funktion darf kein Argument für ein Open Array haben
> > 
> > 
> > This is clearly a Delphi compiler limitation. Should be fixed by [57be441d2](https://github.com/synopse/mORMot2/commit/57be441d2)
> 
> Yes D13 compiles again. D7 Problem is with the default-param overload. Shortest fix i think is to remove on default-param an change some calls of the function. I try to push a fix in my repo, after i finished my current task, if you want to have a look, if not fixed by you in the meanwhile.

Createad a PR: more than expected. Maybe there is a easier solution:

https://github.com/synopse/mORMot2/compare/master...tobfel:mORMot2:fix-d7-winerrorshort

synopse (approved):
Looks great now!

PR #498, opened 2026-07-22
Add my fixes to mormot.lib.win7zip.pas

raelb:
Add a test runner, and docs to the test/win7zip folder


synopse:
Please remove your test program from the pull request.
I guess it would be better integrated in the existing regression tests, in method TTestCoreCompression._7Zip of test.core.data.pas.

synopse (changes requested):
The regression tests should be moved to TTestCoreCompression._7Zip

raelb:
Ok, I will look at it on the weekend

raelb:
Ok, I did it:

**10 new checks** in `_7Zip` itself:

| Tag    | What it verifies                                        |
| ------ | ------------------------------------------------------- |
| `pw1`  | ZipCrypto archive lists exactly 1 entry                 |
| `pw2`  | correct password round-trips the content                |
| `pw3`  | wrong password → boolean `Extract` returns `false`      |
| `pw4`  | wrong password → `RawByteString` overload returns `''`  |
| `pw5`  | wrong password → item extraction raises `E7Zip`         |
| `mhe1` | encrypted-header `.7z` can't be opened without password |
| `mhe2` | ...and can't be opened for update without password      |
| `mhe3` | password-seeded update: archive holds both entries      |
| `mhe4` | original entry survived the in-place update             |
| `mhe5` | newly added entry is correct                            |

Plus **8 incidental assertions** from the helpers: `Run7zExtract`/`Run7zUpdate` each do 2 `Check(VarRecToUtf8IsString(...))` parameter validations, and they're invoked 4 times total (3× extract, 1× update).

So the run count went up by **18 assertions**

synopse (approved):
Some minor tweaks to do but I will review it.

Thanks for the feedback.

PR #500, opened 2026-07-22
core.os: rename ErrorShort procedure overloads to fix Delphi 7 (E2250)

tobfel:
## Problem

The recent `WinErrorShort` / `OsErrorShort` refactoring turned each of them into a **function + procedure overload pair sharing the same name**:

```pascal
function  WinErrorShort(Code: cardinal; NoInt: boolean = false): TShort47; overload;
procedure WinErrorShort(Code: cardinal; var Dest: ShortString; NoInt: boolean = false); overload;
```

**Delphi 7 cannot resolve this overload pair.** At every *procedure* call site (e.g. `mormot.core.os.windows.inc:718` inside `OsErrorShort`, `mormot.core.text.pas`, and the `TTestCoreBase` self-tests) D7 reports:

```
[Error] mormot.core.os.windows.inc(718): E2250 There is no overloaded version of 'WinErrorShort' that can be called with these arguments
```

even though the argument types match the procedure overload exactly. Reordering the two declarations does **not** help (confirmed on D7). Newer Delphi and FPC accept it.

## Fix

Rename the "fill a `var ShortString`" procedure forms so there is no longer a function/procedure name overload, matching the existing `GetErrorShortVar` convention already used in this unit:

- `WinErrorShort(Code; var Dest; NoInt)` → `WinErrorShortVar(...)`
- `OsErrorShort(Code; var Dest; NoInt)` → `OsErrorShortVar(...)`

The **function** forms `WinErrorShort()` / `OsErrorShort()` returning `TShort47` (used as expressions, e.g. `WinErrorShort(err)`) are **unchanged**, so all of those call sites keep working with no change. Updated the internal callers in `os.pas`, `os.windows.inc`, `os.posix.inc`, `text.pas`, and the `TTestCoreBase` self-tests.

## Testing

- **Delphi 7**: the reported `E2250` at `windows.inc:718` (and the other procedure call sites) is gone; builds clean.
- **Delphi 13 (dcc64)**: full `mormot2tests` builds clean (610k lines).
- **FPC**: unaffected by the rename.

### DIFF
--- a/src/core/mormot.core.os.pas
+++ b/src/core/mormot.core.os.pas
@@ -832,11 +832,13 @@ function WinErrorConstant(Code: cardinal): PShortString;
 /// return the error code number, and its regular constant on Windows (if known)
 // - e.g. WinErrorShort(5) = '5 ERROR_ACCESS_DENIED' or
 // WinErrorShort($c00000fd) = 'c00000fd EXCEPTION_STACK_OVERFLOW'
-function WinErrorShort(Code: cardinal; NoInt: boolean = false): TShort47; overload;
+function WinErrorShort(Code: cardinal; NoInt: boolean = false): TShort47;
   {$ifdef FPC} inline; {$endif} // Delphi has sometimes issues inlining this
 
-/// return the error code number, and its regular constant on Windows (if known)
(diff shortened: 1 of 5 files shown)


synopse (approved):
Thanks.

Will merge and review.

PR #504, opened 2026-07-27
fix: valid handle testing under linux 64bits

AntoineGS:
I was running into an issue with log creation under Linux 64bits, it would return an INVALID_HANDLE_VALUE ($ffffffff) which PtrInt() would cast it as 4294967295 instead of -1, thus making it look like it was a success.

### DIFF
--- a/src/core/mormot.core.os.posix.inc
+++ b/src/core/mormot.core.os.posix.inc
@@ -1491,7 +1491,9 @@ end;
 
 function ValidHandle(Handle: THandle): boolean;
 begin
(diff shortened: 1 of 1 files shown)


synopse:
I am not sure THandle is a 32-bit on all POSIX.
On linux, ok.
The FPC doc states it should be stated opaque:
https://www.freepascal.org/docs-html/rtl/system/thandle.html

synopse:
I have made a proper fix I hope.

AntoineGS:
Oh that's clever, I was unaware of the SizeOf() compiler directive, good to know!

Issue #487, opened 2026-06-25
Basic XML parser


synopse:
There is already some XML generation code in the library.

A minimal XML parser should be included in `mormot.core.fmt.pas`.

### 1. One XML simple SAX parser

Reference code for a minimal effective memory-based XML parser:
https://github.com/laoqiuqiu/pixie/blob/master/source/common/Pixie.SimpleXml.pas

This parser should then be used either:
- To reformat/normalize the XML;
- To search for attributes on the fly, perhaps recursively;
- To fill a tree of nodes.

### 2. DOM complex structure - option 1: general nodes hierarchy

- Creating a tree of XML nodes, perhaps with the same API surface as FPC's fcl-xml.
- The input buffer is expected to remain available in memory.
- Allocator-free pattern for the content: should not allocate any string during parsing, but reference to the original buffer, and maybe only allocate RawUtf8 values on demand or when really needed a text value need to be unescaped. No in-place unescape is needed: we keep the unescaped values in-place, and only allocate when escape is needed. 
- Allocator-free pattern for the nodes themself (nice to have): should not allocation any class, but create fake classes as pointer of records, allocated in chunks.
- Resulting code should reuse as many existing code as possible, being fast but not absolutely fastest nor the least memory consumer (16 bytes per nodes could be enough).
- The parser should be able to be run as SAX or as DOM generator.

Reference code for the optional nodes allocation:
https://blog.grijjy.com/2020/10/07/an-xml-dom-with-just-8-bytes-per-node/
https://github.com/neslib/Neslib.Xml

### 3. DOM complex structure - option 2: use variants for nodes

The SAX could be able to generate a custom variant type representation of the original structure.

- The variants would be just be weak references to the original XML document - no allocation, and fast no-op clear.
- Variants could allow late-binding access like `SVGRoot.height` 
- .... perhaps too error prone from the code point of view.

### 4. Investigate recursive SAX

- the SAX parser store positions of each node as integer position into the entry buffer.
- Search for a node and return the position of the matching node;
- Allow attributes search directly from the returned position;
- Convert a value on demand.

OK - investigate further, and start from actual simple needs.

zen010101:
Hi ab,

I would like to take this one, following the same pattern as the YAML support I contributed in #466 (same unit, symmetric API family).

**Actual need and test corpus.** We are migrating a Java + Tomcat admin server to mORMot 2 on embedded aarch64 Linux boxes. Two concrete XML needs:

[...]

PR #507, opened 2026-07-31
fmt: basic XML parser - SAX TXmlParser + XmlToVariant late-binding (#487)

zen010101:
Implements directions 1 + 3 of #487, as proposed in https://github.com/synopse/mORMot2/issues/487#issuecomment-5140101717.

**SAX layer** — `TXmlParser` record in `mormot.core.fmt.pas`: zero-allocation scan over an UTF-8 buffer; `Name`/`Value` point within the input; token positions available to resume a scan (groundwork for direction 4); well-formedness of the tags nesting is verified; `EXmlException` with the faulty line number on malformed input.

"Basic profile" scope, by design:

- no DTD processing at all — which makes the parser immune to entity expansion (billion laughs) by construction; only the five predefined entities and numeric character references are resolved, anything else raises;
- the parser core is UTF-8 only;
- namespaces: the prefix is kept as part of the name (`xpoStripNamespacePrefix` option to strip it), no URI binding;
- `xpoKeepComments` / `xpoKeepPI` / `xpoKeepWhiteSpace` options — those tokens are silently skipped by default;
- non-goals: XPath, XML Schema validation, XSLT.

**Variant layer** — `XmlToVariant` / `TryXmlToVariant` / `XmlToJson`, mirroring the `YamlToVariant()` API family from #466 (same unit): elements as object fields, repeated siblings gathered into arrays, `@` prefix for attributes, `#text` field for mixed content, CDATA as text; all values remain lossless strings, since XML is untyped text by nature.

**Tests**: new `TTestCoreXml` suite (SAX tokens / entities / options / errors + the mapping conventions, 219 assertions); the whole mormot2tests suite is green on FPC i386-win32 (199,376,126 assertions). Also validated against a corpus of real ONVIF device responses (full SOAP envelopes from Dahua/Hikvision cameras: namespaces, deep nesting, repeated profile elements) produced by a gsoap-based stack.

Points I would like your feedback on:

- the exact placement of the variant-layer helpers (kept in the "Basic XML Conversions" section for now);
- the attribute/child mapping convention (`@` / `#text`);
- direction 2 (zero-allocation node tree) can build on this SAX layer as a second step — the token positions are already exposed for it.

### DIFF
--- a/src/core/mormot.core.fmt.pas
+++ b/src/core/mormot.core.fmt.pas
@@ -143,6 +143,120 @@ procedure AddXmlEscape(W: TTextWriter; Text: PUtf8Char);
 function AddJsonToXml(W: TTextWriter; Json: PUtf8Char; ArrayName: PUtf8Char = nil;
   EndOfObject: PUtf8Char = nil): PUtf8Char;
 
+type
+  /// exception raised by TXmlParser on invalid or unsupported XML input
+  EXmlException = class(ESynException);
+
+  /// the kind of tokens returned by TXmlParser.Next
+  // - xtEof is set once the end of the input buffer has been reached
+  // - xtElementStart is returned for each opening <name> or <name ...> tag,
+  // with the parser Name filled - any attribute following as xtAttribute
+  // - xtElementEnd is returned for each </name> or self-closing '/>' mark,
+  // with the parser Name filled
+  // - xtAttribute is returned for each name="value" pair, Name/Value filled
+  // - xtText is returned for any text content between markup, Value filled
+  // with the raw (possibly escaped) text - see ValueToUtf8 to unescape it
+  // - xtCData is returned for each <![CDATA[...]]> section, Value filled
+  // with the verbatim content
+  // - xtComment is returned for each <!--...--> section, if xpoKeepComments
+  // option was defined at Init - silently skipped otherwise
+  // - xtPI is returned for each <?name ...?> processing instruction, if the
+  // xpoKeepPI option was defined at Init - silently skipped otherwise
+  TXmlToken = (
+    xtEof,
+    xtElementStart,
+    xtElementEnd,
+    xtAttribute,
+    xtText,
+    xtCData,
+    xtComment,
+    xtPI);
+
+  /// option to refine TXmlParser process
+  // - xpoStripNamespacePrefix would remove any 'ns:' prefix from the reported
+  // element and attribute names - the prefix is part of the name otherwise
+  // (this "basic" parser does no URI namespace binding by design)
+  // - xpoKeepComments and xpoKeepPI would return xtComment / xtPI tokens,
+  // which are silently skipped by default
+  // - xpoKeepWhiteSpace would return xtText tokens made only of whitespace,
+  // which are silently skipped by default
+  TXmlParserOption = (
+    xpoStripNamespacePrefix,
+    xpoKeepComments,
+    xpoKeepPI,
+    xpoKeepWhiteSpace);
+
+  /// options to refine TXmlParser process
+  TXmlParserOptions = set of TXmlParserOption;
+
+  /// zero-allocation SAX-like parser over an XML UTF-8 memory buffer
+  // - a "basic" parser, from actual simple needs: no DTD support (which makes
+  // it immune to entity expansion attacks by design), only the five XML
+  // predefined entities and numeric character references, prefixes kept as
+  // part of the names with no URI namespace binding
+  // - the input buffer is expected to be UTF-8 encoded and to remain available
+  // in memory during the whole parsing: Name and Value do point within it,
+  // with no memory allocation during the scan - see NameToUtf8/ValueToUtf8
+  // - well-formedness of the tags nesting is verified, and any syntax or
+  // nesting error would raise an EXmlException with the faulty line number
+  // - usage: call Init() then Next in a loop, e.g. as
+  // ! x.Init(pointer(xml), length(xml));
+  // ! while x.Next do
+  // !   case x.Kind of
+  // !     ...
+  {$ifdef USERECORDWITHMETHODS}
+  TXmlParser = record
+  {$else}
+  TXmlParser = object
+  {$endif USERECORDWITHMETHODS}
+  private
+    fCur, fBegin, fAfter, fToken: PUtf8Char;
+    fOptions: TXmlParserOptions;
+    fInElement: boolean; // scanning attributes, up to '>' or '/>'
+    fStack: array of TValuePUtf8Char; // opened elements raw names
+    procedure ParseError(pos: PUtf8Char; const reason: RawUtf8);
+    function ParseName(var p: PUtf8Char): TValuePUtf8Char;
+    procedure SetName(const raw: TValuePUtf8Char);
+  public
+    /// the current token kind, as set by the last Next call
+    Kind: TXmlToken;
+    /// how many elements are currently opened
+    // - incremented after a xtElementStart, decremented after a xtElementEnd
+    Depth: PtrInt;
+    /// the current token name, pointing within the input buffer
+    // - set for xtElementStart, xtElementEnd, xtAttribute and xtPI tokens
+    Name: TValuePUtf8Char;
+    /// the current token raw value, pointing within the input buffer
+    // - set for xtAttribute, xtText, xtCData, xtComment and xtPI tokens
+    // - may still contain XML entities: use ValueToUtf8 to decode them
+    Value: TValuePUtf8Char;
+    /// prepare the parsing of a given XML UTF-8 buffer
+    // - any UTF-8 BOM would be ignored
+    // - the buffer is expected to remain available during the whole parsing
+    procedure Init(Text: PUtf8Char; TextLen: PtrInt;
+      Options: TXmlParserOptions = []);
+    /// iterate to the next token of the input, returning false at xtEof
+    // - raises an EXmlException on any malformed or unsupported input
+    function Next: boolean;
+    /// returns the current Name as an allocated UTF-8 string
+    procedure NameToUtf8(out result: RawUtf8);
+      {$ifdef HASINLINE}inline;{$endif}
+    /// returns the current Value as an allocated UTF-8 string
+    // - decoding any XML entity, unless the current token is a verbatim
+    // xtCData/xtComment section
+    procedure ValueToUtf8(out result: RawUtf8);
+    /// the offset of the current token in the input buffer
+    // - could be used to store a position, then resume a scan from it
+    function Position: PtrInt;
+      {$ifdef HASINLINE}inline;{$endif}
+  end;
+
+/// decode the five XML predefined entities and numeric character references
+// - i.e.   &lt; &gt; &amp; &apos; &quot;   and   &#nnn; &#xhh;   patterns
+// - as used by TXmlParser.ValueToUtf8, or to be called directly
+// - raises EXmlException on any other (i.e. undefined) entity
+procedure XmlUnescape(Text: PUtf8Char; TextLen: PtrInt; out result: RawUtf8);
+
 
 { ************* YAML 1.2 core-schema to JSON or TDocVariant Support }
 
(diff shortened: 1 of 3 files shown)


src/core/mormot.core.fmt.pas
```diff
@@ -1403,6 +1549,616 @@ function JsonToXml(const Json, Header, NameSpace: RawUtf8): RawUtf8;
...
+procedure TXmlParser.ValueToUtf8(out result: RawUtf8);
+begin
+  if (Kind in [xtCData, xtComment]) or
+     (ByteScanIndex(pointer(Value.Text), Value.Len, ord('&')) < 0) then
+    FastSetString(result, Value.Text, Value.Len)
+  else
+    XmlUnescape(Value.Text, Value.Len, result);
+end;
+
+procedure XmlUnescape(Text: PUtf8Char; TextLen: PtrInt; out result: RawUtf8);
```
synopse, line 2056: Perhaps a dedicated procedure to output not as RawUtf8 but within a TTextWriter instance.
Something similar to AddHtmlUnescape().
In fact, I just saw that AddHtmlUnescape() does not decode numeric character references.
So a function (inlined) dedicated to numeric character references could be a good idea - similar to EntityToUcs4(), and reuse for both AddHtmlUnescape() and AddXmlUnescape() which has only 5 entitites.

zen010101:
Platform validation status:

| target | result |
|---|---|
| FPC 3.2.3 i386-win32 | whole suite green (199,376,126 assertions) |
| FPC 3.2.3 aarch64-linux (Rockchip RK3588 hardware) | whole suite green (198,860,260 assertions) |
| Delphi 37.0 Win32 | TTestCoreXml green (219 assertions) |
| Delphi 37.0 Win64 | TTestCoreXml green (219 assertions) |

The Delphi run caught a string-literal encoding issue in the new test itself (a `#$e4#$b8#$ad` UTF-8 literal is re-encoded from UTF-16 chars by the Delphi compiler) — fixed by computing the expected CJK text at runtime via `Ucs4ToUtf8()`. Parser code unchanged.

src/core/mormot.core.fmt.pas
```diff
@@ -1403,6 +1549,616 @@ function JsonToXml(const Json, Header, NameSpace: RawUtf8): RawUtf8;
...
+        inc(line);
+      inc(p);
+    end;
+  EXmlException.RaiseUtf8('XML error at line %: %', [line, reason]);
+end;
+
+function TXmlParser.ParseName(var p: PUtf8Char): TValuePUtf8Char;
+var
+  s: PUtf8Char;
+begin
```
synopse, line 1658: revert s and p logic in the loop: so that the inc(s) would be on a local variable (i.e. a register) whereas inc(p) would increment a pointer by reference so would be slower.

src/core/mormot.core.fmt.pas
```diff
@@ -1403,6 +1549,616 @@ function JsonToXml(const Json, Header, NameSpace: RawUtf8): RawUtf8;
...
+        // decode one of the five predefined entities
+        ch := #0;
+        case s - a of
+          2:
+            if PWord(a)^ = ord('l') + ord('t') shl 8 then
+              ch := '<'
+            else if PWord(a)^ = ord('g') + ord('t') shl 8 then
+              ch := '>';
+          3:
+            if CompareMemSmall(a, PAnsiChar('amp'), 3) then
```
synopse: Use PCardinal(a)^ and $00ffffff here

src/core/mormot.core.fmt.pas
```diff
@@ -1403,6 +1549,616 @@ function JsonToXml(const Json, Header, NameSpace: RawUtf8): RawUtf8;
...
+      if i < 0 then
+      begin
+        W.AddNoJsonEscape(p, e - p);
+        break;
+      end;
+      W.AddNoJsonEscape(p, i);
+      amp := p + i;
+      a := amp + 1; // just after '&'
+      s := a;
+      while (s < e) and
```
synopse: it would be faster not to compute the length first, but use a cascaded set of case ... of.
I could do that myself.

src/core/mormot.core.fmt.pas
```diff
@@ -143,6 +143,152 @@ procedure AddXmlEscape(W: TTextWriter; Text: PUtf8Char);
...
+  {$ifdef USERECORDWITHMETHODS}
+  TXmlParser = record
+  {$else}
+  TXmlParser = object
+  {$endif USERECORDWITHMETHODS}
+  private
+    fCur, fBegin, fAfter, fToken: PUtf8Char;
+    fOptions: TXmlParserOptions;
+    fInElement: boolean; // scanning attributes, up to '>' or '/>'
+    fStack: array of TValuePUtf8Char; // opened elements raw names
```
synopse: Could we use a static allocated stack here like for JSON?
The parser is optimized for the common SAX use case. Instead of storing `TValuePUtf8Char` (16 bytes on 64-bit), each stack entry packs a 25-bit offset from `fBegin` (up to 32 MiB input) and a 7-bit name length (up to 127 UTF-8 bytes) into a single `UInt32`. XML names longer than 127 bytes are legal but exceedingly rare in practice, so this keeps the stack compact and cache-friendly without impacting real-world documents. A fixed stack depth of 256 (1 KiB total) is already far beyond typical XML nesting while still providing protection against pathological inputs.
Another refinment: when the nesting level returns to 0, we've closed the root element, so no open element names still need to refer back into the earlier part of the buffer. At that point we can safely reset the origin (fBegin or the packed offset base) to the current parsing position. So no 32MiB limit any more.

src/core/mormot.core.fmt.pas
```diff
@@ -1403,6 +1549,616 @@ function JsonToXml(const Json, Header, NameSpace: RawUtf8): RawUtf8;
...
+
+procedure TXmlParser.NameToUtf8(out result: RawUtf8);
+begin
+  FastSetString(result, Name.Text, Name.Len);
+end;
+
+procedure TXmlParser.ValueToUtf8(out result: RawUtf8);
+begin
+  if (Kind in [xtCData, xtComment]) or
+     (ByteScanIndex(pointer(Value.Text), Value.Len, ord('&')) < 0) then
```
synopse: nice optimization to not unescape when not needed. the most common case for sure.  ;)

src/core/mormot.core.fmt.pas
```diff
@@ -1403,6 +1549,616 @@ function JsonToXml(const Json, Header, NameSpace: RawUtf8): RawUtf8;
...
+  p, e, s, a, amp: PUtf8Char;
+  i: PtrInt;
+  c: cardinal;
+  ch: AnsiChar;
+  ent: RawUtf8;
+  buf: array[0..7] of AnsiChar;
+begin
+  p := Text;
+  if (p = nil) or
+     (ByteScanIndex(pointer(p), TextLen, ord('&')) < 0) then
```
synopse: could have been done by caller.
another reason to split into sub functions.
this is the purpose of the `amp` parameter in AddHtmlUnescape().

src/core/mormot.core.fmt.pas
```diff
@@ -1403,6 +1549,616 @@ function JsonToXml(const Json, Header, NameSpace: RawUtf8): RawUtf8;
...
+      while (s < e) and
+            (s^ <> ';') and
+            (s - a < 12) do
+        inc(s);
+      if (s = e) or
+         (s^ <> ';') then
+        s := nil
+      else if a^ = '#' then
+      begin
+        // decode &#nnn; or &#xhh; numeric character reference
```
synopse: sub-function NumCharToUcs4()

src/core/mormot.core.fmt.pas
```diff
@@ -1403,6 +1549,616 @@ function JsonToXml(const Json, Header, NameSpace: RawUtf8): RawUtf8;
...
+              s := nil;
+              break;
+            end;
+        if (s <> nil) and
+           ((c = 0) or
+            ((c >= $d800) and
+             (c <= $dfff))) then
+          s := nil; // reject #0 and surrogates
+        if s <> nil then
+          W.AddNoJsonEscape(@buf, Ucs4ToUtf8(c, @buf));
```
synopse: use TTextWriter.AddUcs4() with no temp buffer.

test/test.core.data.pas
```diff
@@ -9137,6 +9157,206 @@ procedure TTestCoreYaml.OpenapiEquivalence;
 end;
 
 
+{ TTestCoreXml }
```
synopse, line 9160: nice to see good testing coverage.

synopse (changes requested):

synopse:
Nice work!

I have made some comments.

Thanks for the feedback!

synopse:
If you are a bit lost with my comments, I could just merge it and make the changes myself.

zen010101:
Thanks for the detailed review — not lost at all :) Addressing all of it now: TTextWriter-based `AddXmlUnescape()` with a shared `NumCharToUcs4()` helper also wired into `AddHtmlUnescape()`, the register-friendly scan loops, the `PCardinal`/cascaded-case entity matching, `TTextWriter.AddUcs4()`, and the packed UInt32 static stack with the origin-reset trick to lift the 32 MiB limit — nice design, implementing it as specified. Will push the rework later today.

zen010101:
All remarks addressed in cc261e39:

- `AddXmlUnescape(W: TTextWriter)` with `XmlUnescape()` as a thin wrapper; the five entities matched via a cascaded case + `PCardinal` comparisons
- shared `NumCharToUcs4()` — also wired into `AddHtmlUnescape()`, which never decoded `&#nnn;`/`&#xhh;` references, and its output now goes through `TTextWriter.AddUcs4()` (so code points above the BMP render properly too)
- the `TXmlParser` stack is now the static 1KB packed array you sketched (25-bit offset + 7-bit length, origin reset when the nesting returns to 0 — so no 32 MB limit across consecutive documents, only within a single root)
- register-friendly `ParseName()` scan
- new tests: HTML numeric references, dangling `&`, 256+ nesting and 127+ names rejection

Re-validated everywhere: whole suite green on FPC i386-win32 (199.5M assertions) and FPC aarch64-linux on RK3588 hardware (199.0M), `TTestCoreXml` green on Delphi 37.0 Win32/Win64 (223 assertions each).

synopse (approved):
Sounds very fine now.

zen010101:
Ran a differential check of the refactored parser against the pre-merge version (same test corpus, plus some boundary cases we had not covered). Five regressions, all reproducible.

**1. Out-of-bounds read at the end of the buffer** — `mormot.core.fmt.pas:1920`, `if p^ = '<' then` with no `p < e` guard: the whitespace-probing loop can leave `p = e`. With a buffer whose last byte sits exactly on a committed page boundary (VirtualAlloc, next page unmapped) this is an access violation. Enough to trigger it: `'<a/>'#10`, `'<a>x</a> '`, or any all-whitespace input — i.e. any XML file ending with a line feed. A `RawUtf8` has its implicit #0 so it survives, but an mmap'd file or a network buffer slice does not. Without the page boundary it silently yields one extra empty `xtText` token.

**2. Leading whitespace of mixed content is now dropped** — same lines. The probe loop and the returned text start share `p`, so the skipped blanks are lost. It is asymmetric, which is what makes it look unintended:

```
                         before            after
<a>  hello</a>           "  hello"         "hello"
<p><b>Hi</b> there</p>   #text=" there"    #text="there"
<a> both </a>            " both "          "both "
```

Both 1 and 2 go away by probing on a copy (as the previous `s` variable did) and adding the `p < e` guard.

**3. 32-bit wraparound in `NumCharToUcs4` lets entities smuggle through** — `:1258-1275`: moving the range check out of the digit loop means `c` (a `cardinal`) wraps, and the wrapped value can land back in the valid range:

```
&#x10000003C;  ->  '<'
&#x100000026;  ->  '&'
&#4294967361;  ->  'A'
```

This one also reaches HTML, since `AddHtmlUnescape` shares the helper: `HtmlUnescape('x &#x10000003C; y')` now returns `x < y`. Keeping one `if c > UNICODE_MAX then exit` inside the loop (or accumulating into a `QWord`) restores it.

**4. CDATA/PI ending exactly at the buffer end is reported as unfinished** — `:1849` `if p >= e` after `dec(e, 3)` (and `:1870` after `dec(e, 2)`): `p = e` is the valid case. `'<a></a><![CDATA[x]]>'` and `'<a/><?pi x?>'` now raise. `>=` should be `>`.

Also, two diagnostics Delphi emits on the new code and FPC does not: `:1871` `W1036 Variable 's' might not have been initialized` — the PI branch lost its `s := p` but still passes `s` to `ParseError`, so the reported line number is wrong (line 1 instead of 6 for a PI truncated further down); and `:1851` `H2077 Value assigned to 'e' never used`, which is the same spot as regression 4.

**5. The nesting limit lost one level** — `:1897`, `Depth = high(fStack)`. `Depth` is the pre-increment index and `fStack: array[byte]` has 256 slots, so the previous `Depth > high(fStack)` refused the 257th start tag and slot 256 was never written: a legal 256-level document now raises `too much nesting`. I measured it before saying so — canaries right after `fStack` (a sentinel field at that exact offset) stay intact at 254/255/256/257/300 levels on the pre-merge version, and 256 levels parse fine there.

All of these are small enough that you may prefer to fix them in place while the code is fresh; say the word if you would rather have a PR. On my side I will add the coverage that missed them: leading whitespace in mixed content, CDATA/PI at end of buffer, oversized numeric references, the 256/257 depth boundary, and a non-`RawUtf8` buffer end.

On the TODO about returning an error code instead of raising: sounds right to me for the SAX layer. If you go that way, keeping `TryXmlToVariant` as the no-exception entry at the variant level would keep the pair symmetric with the YAML functions.

synopse:
I tried to fix the regressions.

About Depth: byte and the last high(fStack) it is on purpose - I change the type to byte for better alignment of internal fields and more efficient assembly code generation. Reducing the max depth from 256 to 255 does not hurt.

Perhaps some more tests to add to avoid such regressions in any future refactoring? ;)

zen010101:
Re-ran the checks against 3c087a6e5. Three of the five are gone (leading
whitespace, the numeric overflow, CDATA at end of buffer). Two corrections
on my side, and one still open.

**Corrected: the nesting limit.** You are right, and my earlier message was
wrong. I had missed that you also changed `Depth` to `byte` in the same
refactoring. With that type `Depth > high(fStack)` can never be true and
`inc(Depth)` wraps at 255, so `Depth = high(fStack)` is the only correct
test. 255 levels is plenty. Sorry for the noise.

**Corrected: the out-of-bounds read.** I mis-attributed it. The fault is in
`Init()`, not in `Next()`: the BOM check reads `PCardinal(Text)^` while only
requiring `TextLen >= 3`. So a 3-byte buffer whose last byte sits on the last
byte of a committed page faults before parsing even starts — and that is my
own code from the original PR, identical in both versions. Measured on a
`VirtualAlloc` guard page:

```
TextLen=1  "x"      Init ok
TextLen=2  "xy"     Init ok
TextLen=3  "   "    Access violation      <-- both 3c087a6e5 and cc261e395
TextLen=3  "<a>"    Access violation
TextLen=4  "xyzw"   Init ok
```

`(TextLen >= 4) or a 3-byte compare` would fix it. Happy to send that as a
one-liner if you prefer.

**Still open: the PI branch of regression 4.** You fixed the CDATA side
(`p > e`) but `:1990` still has `if p >= e` after `dec(e, 2)`, so a processing
instruction that ends exactly at the end of the buffer is reported as
unfinished. It needs at least one trailing byte to parse:

```
<a/><?pi x?>        ERR unfinished processing instruction
<a/><?pi x?>#10     ok
<a/><?pi x?><b/>    ok
<a/><?pi?>          ERR unfinished processing instruction
```

The comment branch is fine, since it kept the `e - p >= 3` form.

**One behaviour change worth pinning down.** `xpoStripNamespacePrefix` now
strips up to the *last* colon, where the original stripped up to the first:
`<a:b:c/>` reports `c` instead of `b:c`. Yours is closer to the QName local
part, so I would keep it — just noting it so a test locks it in rather than
leaving it to the next refactoring.

On the tests: yes, gladly. I have a `SaxBoundaries` test here (~8k
assertions) that covers exactly these classes — leading/trailing whitespace in
mixed content, CDATA/PI/comment ending at the buffer end, oversized and
malformed numeric references, the depth boundary at 254/255/256, names at
127/128 bytes, namespace stripping with several colons, and a buffer whose
last byte is on a guard page so an over-read faults instead of passing
silently. I mutation-checked it against the four real regressions above plus
the `Init` one, and each is caught. Want me to send it as a PR against
current master?

PR #506, opened 2026-07-31
mvc: add TMvcViewsMustache.OnTranslate property

zen010101:
Wire the existing Mustache `{{"text}}` translate channel for MVC apps: `TMvcViewsMustache` gets an `OnTranslate: TOnStringTranslate` property, passed down to `TSynMustache.Render()`.

The translation callback types, the `{{"text}}` tag and the `Render()` parameter were already all in place — the MVC layer just never passed the callback down, and exposed no property for it.

Language negotiation is deliberately left to the application (e.g. via a threadvar set at each request start), as the minimal contract.

Includes a `TTestServiceOrientedArchitecture.MvcViewsMustache` test; the whole SOA suite passes (35.5M assertions, FPC i386-win32).

Discussion: https://synopse.info/forum/viewtopic.php?id=7592 — this is the first step (PR-A) of the basic i18n proposal there; the `mormot.core.i18n` unit (PR-B) will follow the forum discussion.

PR #510, opened 2026-08-02
core: add mormot.core.i18n - basic i18n unit

zen010101:
Follows the discussion in https://synopse.info/forum/viewtopic.php?id=7592 — this is PR-B of that proposal, PR #506 having wired the MVC side.

**`TSynLanguage`** — one translation table per language, keyed on the original English text, so a missing key naturally falls back to it. Loaders: `.po` as the main format (its msgid *is* the English text, and translators get the whole gettext tooling), `.ini` / `.msg` on top of the existing `FindSectionFirstLine`, `.yaml` and `.json` with its relaxed JSON5 / JSONC / HJson variants — those two came almost for free from our own parsers. `AddFromFile` dispatches on the file extension.

**`TSynLanguages`** — a registry indexed by the existing `TLanguage` enum (not duplicated), with per-thread language selection for web use, `LoadFromFolder` over `<iso>.<ext>` files, and `LoadedLanguages` to fill a language selection list.

**Three wiring channels**, all fed by the same table:

- the Mustache translate tag, via `TMvcViewsMustache.OnTranslate` from #506;
- the `LoadResStringTranslate` slot consumed by `GetCaptionFrom*`;
- on FPC only, the whole executable resourcestring table, through the `objpas.SetResourceStrings()` official API — the per-unit writable table is the only official hook there (the enumeration-style API is gone in 3.x). Delphi keeps its resourcestrings in the executable resources with no such writable runtime table, so `TranslateResourceStrings` is a documented no-op there.

Per-language date/time patterns are rendered through our own `TFormatSettings`, so the `/` and `:` pattern characters are no longer rewritten by the process-wide locale separators — that was silently defeating the whole point of a per-language pattern on a POSIX-locale Linux box.

Deliberately out of scope: VCL/LCL form translation (the v1 part), plurals, and msgctxt.

**Tests**: new `TTestCoreI18n` suite — 201 assertions on FPC, 192 on Delphi (the difference being the FPC-only resourcestring part). Green on FPC 3.2.3 i386-win32, FPC 3.2.3 aarch64-linux on Rockchip RK3588 hardware, Delphi 37.0 Win32 and Win64.

The accented-translation coverage comes from @flydev-fr: the original assertions were all pure ASCII, so the multi-byte UTF-8 path through the `.po` and `.json` loaders was never actually exercised. The branch has since been rebased onto current master, so the diff is purely additive against it.

### DIFF
--- a/src/core/mormot.core.i18n.pas
+++ b/src/core/mormot.core.i18n.pas
@@ -1,0 +1,426 @@
+/// Framework Core Basic Internationalization Support
+// - this unit is a part of the Open Source Synopse mORMot framework 2,
+// licensed under a MPL/GPL/LGPL three license - see LICENSE.md
+unit mormot.core.i18n;
+
+{
+  *****************************************************************************
+
+   Basic Internationalization (i18n) Support
+    - TSynLanguage per-language translation table
+    - TSynLanguages registry with per-thread language selection
+    - global wiring of the framework translation hooks
+
+   Connects the translation slots kept since mORMot 1 mORMoti18n.pas:
+   TOnStringTranslate / TOnUtf8Translate callbacks, LoadResStringTranslate,
+   i18nDateText / i18nDateTimeText, and the TSynMustache {{"text}} channel -
+   see https://synopse.info/forum/viewtopic.php?id=7592
+
+  *****************************************************************************
+}
+
+interface
+
+{$I ..\mormot.defines.inc}
+
+uses
+  sysutils,
+  classes,
+  mormot.core.base,
+  mormot.core.os,
+  mormot.core.unicode,
+  mormot.core.text,
+  mormot.core.datetime,
+  mormot.core.data,
+  mormot.core.variants,
+  mormot.core.json;
+
+
+{ ************* TSynLanguage per-language Translation Table }
+
+type
+  /// exception raised on invalid i18n definition or process
+  EI18nException = class(ESynException);
+
+  /// implement one translation table for a given language
+  // - table keys are expected to be the original English text, following the
+  // mORMot 1 mORMoti18n.pas semantic and the Mustache {{"English text}}
+  // convention - so a missing key just fallbacks to the original English
+  // - thread-safe access via an internal TSynDictionary
+  TSynLanguage = class(TSynPersistent)
+  protected
+    fLanguage: TLanguage;
+    fIso: RawUtf8;
+    fTexts: TSynDictionary; // RawUtf8 (english/key) -> RawUtf8 (translated)
+    fDateFormat, fDateTimeFormat: string;
+  public
+    /// initialize the table for a given language
+    constructor Create(aLanguage: TLanguage); reintroduce;
+    /// finalize this instance
+    destructor Destroy; override;
+    /// merge translations from a JSON object, e.g. {"Hello":"Bonjour"}
+    // - returns the number of added or replaced pairs, or -1 on invalid JSON
+    function AddFromJson(const Json: RawUtf8): integer;
+    /// merge translations from an UTF-8 JSON object file (BOM tolerated)
+    function AddFromJsonFile(const FileName: TFileName): integer;
+    /// translate the supplied text in-place
+    // - returns true if the key was found and Text was replaced
+    // - returns false and leaves Text untouched (fallback to original text)
+    function Translate(var Text: RawUtf8): boolean;
+    /// TOnStringTranslate-compatible callback, e.g. to be assigned to
+    // TMvcViewsMustache.OnTranslate or supplied to TSynMustache.Render()
+    procedure TranslateString(var English: string);
+    /// TOnUtf8Translate-compatible callback
+    // - Translated is left '' if the key was not found (i.e. caller fallback)
+    procedure TranslateUtf8(English: PUtf8Char; EnglishLen: integer;
+      var Translated: RawUtf8);
+    /// the associated language of this table
+    property Language: TLanguage
+      read fLanguage;
+    /// the ISO 639-1 text of this language, e.g. 'en' or 'zh'
+    property Iso: RawUtf8
+      read fIso;
+    /// optional FormatDateTime() pattern used by the i18nDateText hook
+    // - default '' will use the plain RTL DateToStr() rendering
+    property DateFormat: string
+      read fDateFormat write fDateFormat;
+    /// optional FormatDateTime() pattern used by the i18nDateTimeText hook
+    // - default '' will use the plain RTL DateTimeToStr() rendering
+    property DateTimeFormat: string
+      read fDateTimeFormat write fDateTimeFormat;
+    /// how many translation pairs are stored in this table
+    function Count: integer;
+  end;
+
+
+{ ************* TSynLanguages Registry with per-thread Language }
+
+  /// registry of TSynLanguage tables with per-thread language selection
+  // - typical web usage: load the tables once at startup, then call
+  // TSynLanguages.SetThreadLanguage() at each request start (e.g. from an
+  // URI parameter or a cookie), and assign TranslateString to the Mustache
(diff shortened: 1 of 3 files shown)


flydev-fr:
Thanks @zen010101 - I sent smalls modifs on the TTestCoreI18n, see there  https://github.com/zen010101/mORMot2/pull/2#issue-5044342749

zen010101:
Thanks a lot for the patch — merged into the branch as-is, with your authorship preserved.

Your tests cover a real gap: all of my original assertions were pure ASCII, so the multi-byte UTF-8 path through the `.po` and `.json` loaders was never actually exercised. The `WinAnsiToUtf8` + `initialization` approach also keeps the source 7-bit clean, which matches what the rest of the test suite already does.

One detail I appreciated: the explicit `RawUtf8()` casts. `CheckEqual` has a `(pointer, pointer, RawUtf8)` overload in scope, and a plain string literal in the wrong position silently binds to it — I hit exactly that access violation while writing the XML tests, so the casts are not merely cosmetic here.

Verified on two targets, same numbers on both: 201 assertions / 0 failed for the i18n tests (i386-win32 and aarch64-linux). Without your commit `IniAndFiles` reports 93 assertions, with it 102, so the accented cases are genuinely running.

The branch was 26 commits behind and had become conflicting, so I rebased it onto current master before merging your work — the diff is now purely additive against master. Two small non-blocking thoughts, entirely up to you: names like `_eacute` might read a little easier than `_uE9`, and the repeated `RawUtf8('Fa') + _uE7 + 'ade'` expressions could be hoisted into locals. Happy to leave both as they are.

flydev-fr:
Super, thanks for all the effort you put into this :)

synopse:
I guess .mo file format should be somehow supported in addition to .po ascii source file.

zen010101:
Done — .mo is now supported, in addition to .po.

`TSynLanguage.AddFromMo()` / `AddFromMoFile()` are the binary counterpart of `AddFromPo()`: same table, same non-goals. Both endianness are handled, the format revision is checked, and the whole input is validated before anything is merged — so an invalid .mo is rejected as a whole, and never leaves a half-filled table behind. The strings are assumed to be UTF-8, exactly as `AddFromPo()` already does with its source.

The gettext semantics line up on their own: `msgfmt` does not put fuzzy entries in its output unless `--use-fuzzy` is given, and the plural / msgctxt entries are recognizable by their #0 and #4 separators, so they are skipped just like their .po counterparts.

One side effect worth mentioning: `LoadFromFolder()` identifies a language by the file base name, so `fr.po` and `fr.mo` are both loaded and merged. Since a .mo is a compiled artifact of its .po, the two easily fall out of sync — and until now, the winner of a duplicated key depended on the OS folder enumeration order. The files are now sorted before being merged (by name, then by ascending `LANGUAGE_EXT[]` index), so the result is deterministic on every platform, and a compiled `fr.mo` consistently wins over its `fr.po` source. Happy to flip that precedence if you prefer the source to win.

`TTestCoreI18n.MoFormat` adds 167 assertions: the reference sample in both endianness, the header / untranslated / plural / msgctxt skips, the UTF-8 passthrough, an invalid magic, an unsupported major revision, an inflated entry count, and every possible truncation of a valid .mo (each one must be rejected and merge nothing). Same numbers on both targets — 368 assertions / 0 failed for the whole test case, on i386-win32 and x86_64-linux.

The branch is rebased onto current master, so the diff is still purely additive.

synopse (approved):
Seems just fine!

Thanks for the input.

synopse:
In fact, I guess the current Mustache translation process is somewhat wrong:
- it uses transient variables (string or utf-8) for no real benefit
- it does not follow the HTML escape policy

I have fixed it as such:
https://github.com/synopse/mORMot2/commit/93b8d92a6

zen010101:
Agreed on both points - the fix looks right to me.

Escaping the {{"..."}} translation output by default is the correct policy: it makes a
translated block behave exactly like a normal {{variable}} (escaped unless the {{{...}}} / &
invert form is used). That's both safer - a translation that happens to contain <, >, & or "
can no longer inject markup - and less surprising than the previous raw output. For typical
i18n payloads (plain UI labels and messages) the escape is a no-op, and the rare translation
that intentionally carries HTML can opt out via the unescaped form, exactly like any other
Mustache value.

The transient-variable removal is a nice cleanup too: a single pointer with FastAssignNew
avoids the two managed-string temporaries per translated block.

Only thing I'd suggest is a short release-note line, since it is technically a behavior change:
anyone whose translations previously relied on raw HTML output will now get it escaped and
should move those to the unescaped form.

Thanks for spotting and fixing it.

synopse:
Also the tests were broken on Delphi 7/2007: AnsiString are not able to store the "tea" ideogram.

synopse:
1) I have still a logical doubt with SetGlobal and Current: TLanguageFile.
My guess is that SetThreadLanguage() should perhaps NOT affect SetGlobal and its associated functions.
But I also understand the current behavior, with its own threadvar behavior.
It should properly documented, and also about the fact that TranslateResourceStrings is global and not per-thread.

2) I don't understand why _I18nFormat is a global variable.
It should better be part stored as TFormatSettings field in TLanguageFile too, togetehr with DateTimeFormat/DateFormat.

synopse:
I have written a blog article about this
https://blog.synopse.info/?post/2026/08/12/A-New-Native-i18n-Layer-for-mORMot-2

Thanks again @zen010101  for the very nice proposal and Pull Request!

**Any feedback is welcome, as usual!**

zen010101:
Both are fair points, and I'm happy with how you resolved them.

**1) SetGlobal / per-thread language**

The per-thread resolution is intentional -- one process serving concurrent
requests, each rendered in its own language -- so keeping it is right. The
documentation you added (the SetGlobal warning and the reworded
"SetThreadLanguage() or DefaultLanguage" comments) is exactly what was missing,
and the point that TranslateResourceStrings is unavoidably process-wide is well
taken.

**2) _I18nFormat global vs TLanguageFile.DateTimeSettings**

Agreed -- moving it into the table as DateTimeSettings is the better design, and
I would keep your version.

For the history: the global was a deliberately minimal fix for one specific bug
-- FormatDateTime rewrites the '/' and ':' of the pattern into the
DateSeparator / TimeSeparator of the supplied settings, which default to the
process locale globals, so on a POSIX/C-locale box 'yyyy/mm/dd' silently
rendered as '2026-08-13'. Since the only thing I varied per language was the
pattern string, one neutral settings block (with '/' and ':' forced literal)
looked sufficient.

The downside I under-scoped is exactly what your change fixes: sharing one
settings block froze not only the separators but also the month/day names and
AM/PM to the RTL locale for every language, so an 'mmmm' pattern in a French
table would render "January" rather than "janvier" -- wrong for a translation
unit. A per-language TFormatSettings lets each language carry its own separators
and names, while defaulting to the neutral '/'-':' block keeps the original
locale-rewriting fix intact. Nothing to defend on my side here.

(And thanks for the Delphi 7/2007 fix -- the "tea" ideogram not fitting in an
AnsiString is a good catch.)

zen010101:
> I have written a blog article about this
> https://blog.synopse.info/?post/2026/08/12/A-New-Native-i18n-Layer-for-mORMot-2
>
> **Any feedback is welcome, as usual!**

Thank you -- the write-up reads really well, and the .m18n compact/embeddable
format is a nice touch that takes it well past the original proposal. Credit is
shared: your rework (the thread-safe tables, the per-language DateTimeSettings,
the FPC resourcestring path) and flydev's accented-translation tests made it
considerably more complete than what I first sent. It was a pleasure to
contribute -- thanks for driving it into the core.

synopse:
The last missing feature is probably TForm translation with VCL, FMX and LCL... :)

flydev-fr:
I see exactly what you mean

synopse:
I have made a first draft.
https://github.com/synopse/mORMot2/issues/539

PR #511, opened 2026-08-02
TXmlParser input-boundary tests, and 3 defects they surfaced

zen010101:
Follow-up to #507, as offered in https://github.com/synopse/mORMot2/pull/507#issuecomment-5159805902.
Rebased on current master (3c087a6). Thanks for the review and the polishing you did on that one.

### The test: `TTestCoreXml.SaxBoundaries` (+328 lines in `test/test.core.data.pas`)

It covers the input-boundary classes the existing SAX tests do not reach:

- **markup whose last byte is the last byte of the buffer** - CDATA, PI, comment,
  element, text - each paired with its truncated form, so relaxing an
  end-of-buffer test cannot silently start accepting malformed input
  (11 `ExpectRaise` cases);
- **numeric character references** at `&#x10FFFF;` (must decode to the expected
  4 bytes), above the Unicode range, and crafted to wrap a 32-bit accumulator;
- **the nesting limit at its exact value**: 253..257 levels of well-formed
  elements, with per-level distinct names and with a shared name, plus the
  origin reset between sibling roots, plus 127/128-bytes names;
- **buffers with no `#0` terminator and no readable byte past their end**.
  On Windows the payload is flushed against a `PAGE_NOACCESS` page, so a
  single-byte overread raises an access violation instead of going unnoticed;
  elsewhere an exact-size `GetMem()` block is used, which ASAN or valgrind
  catch just as well.

### Three defects it surfaced, fixed in the same commit (`mormot.core.fmt.pas`, 16 lines)

1. `Init()` tested the leading UTF-8 BOM with a 4-bytes `PCardinal` read while
   only requiring `TextLen >= 3`, so a 3-bytes input ending on a page boundary
   faulted before parsing even started. Now compares 2+1 bytes through
   `BOM_UTF8`, as `BomFile()` already does.

2. In current master the PI branch tests `if p >= e` after `dec(e, 2)`, whereas
   the CDATA branch tests `p > e` - this is the PI half of regression 4 in #507.
   So `<a/><?pi x?>` raises "unfinished processing instruction", while
   `<a/><?pi x?><b/>` parses fine. The patch aligns the PI branch on the CDATA
   one, and the truncation cases above show nothing else opens up.

3. With `xpoKeepPI`, a PI whose body is only blanks (`<?pi   ?>`) walked `p[-1]`
   backwards with no lower bound - `ParseName()` had already consumed those
   blanks, so `fCur` could point straight at the `?`. `Value.Len` came out
   **negative**: -3 for that input, -1 for `<?pi ?>`. This one is independent of
   the #507 regressions.

### Mutation-checked

Reverting any single fix makes the new test fail or fault:

| reverted                                     | detected as                  |
|----------------------------------------------|------------------------------|
| `Init` BOM back to `PCardinal`               | access violation (guard page)|
| PI end back to `p < e` / `p >= e`            | 4 failed assertions          |
| PI trim lower bound removed                  | access violation             |
| #507 r2 `p := fToken` rewind                 | 6 failed assertions          |
| #507 r3 `c > UNICODE_MAX` check              | access violation             |
| #507 r4 CDATA `p < e` / `p >= e`             | 4 failed assertions          |
| `Depth = high(fStack)` back to `>`           | access violation             |

The last row is the empirical form of what I withdrew in #507: with `Depth: byte`,
`Depth > high(fStack)` is unsatisfiable, so `Depth = high(fStack)` is the only
test that holds - and the limit is therefore 255 levels, not 256.

### Tested

- FPC i386-win32: `TOTAL assertions=577 failed=0` (`SaxBoundaries` contributes 354)
- FPC aarch64-linux, on device: `TOTAL assertions=543 failed=0` (320) - the delta
  is the guard-page assertions, which are Windows-only

### DIFF
--- a/src/core/mormot.core.fmt.pas
+++ b/src/core/mormot.core.fmt.pas
@@ -1791,7 +1791,8 @@ procedure TXmlParser.Init(Text: PUtf8Char; TextLen: PtrInt;
 begin
   if (Text <> nil) and
      (TextLen >= 3) and
-     (PCardinal(Text)^ and $00ffffff = $00bfbbef) then
+     (PWord(Text)^ = BOM_UTF8 and $ffff) and       // no PCardinal 4-bytes read
+     (PByteArray(Text)[2] = BOM_UTF8 shr 16) then  // on a 3-bytes-only buffer
   begin
     inc(Text, 3); // ignore any UTF-8 BOM
     dec(TextLen, 3);
@@ -1980,22 +1981,23 @@ begin
             // <?name ...?> processing instruction
             inc(p);
             p := ParseName(p, e, 'void or invalid PI name');
-            fCur := p;
+            fCur := p; // just after the name and its trailing blanks
             dec(e, 2);
-            while (p < e) and
+            while (p <= e) and
                   ((p^ <> '?') or
                    (p[1] <> '>')) do
               inc(p);
-            if p >= e then
+            if p > e then
               ParseError(Name.Text, 'unfinished processing instruction');
             if xpoKeepPI in Options then
             begin
               Value.Text := fCur;
-              while p[-1] <= ' ' do
-                dec(p);
+              while (p > fCur) and
+                    (p[-1] <= ' ') do
+                dec(p); // trim trailing blanks, but never before the value
               Value.Len := p - Value.Text;
               while p^ <= ' ' do
-                inc(p);
+                inc(p); // reach back the '?>' ending
               inc(p, 2);
               Kind := xtPI;
               break;
--- a/test/test.core.data.pas
+++ b/test/test.core.data.pas
@@ -211,6 +211,11 @@ type
     procedure SaxOptions;
     /// malformed input and the "basic profile" rejection set (e.g. DTD)
     procedure SaxErrors;
+    /// buffer/nesting/entity boundaries which no other test does cover
+    // - markup ending at the very last byte of the input, oversized numeric
+    // references, the 255 nesting levels limit, 127-bytes names, and input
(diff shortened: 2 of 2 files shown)


zen010101:
Rebased on `bc18941` - no conflict, and the redundant `inc(e, 2)` this branch
was also removing is now covered by your cleaning, so the diff got smaller.

One interaction seemed worth checking rather than eyeballing: `bc18941` moves
the PI `inc(e, 2)` onto the `continue` path, while this branch relaxes the same
branch to `p <= e` / `p > e`. The two are orthogonal, but only if `e` stays
balanced across a skipped PI. Verified on the rebased build, without
`xpoKeepPI` so the `continue` path is taken: `<?pi x?><a/>`,
`<r><?pi x?>tail</r>`, `<r><?pi x?><![CDATA[z]]></r>`, `<?pi a?><?pi b?><a/>`,
`<?pi?><a/>`, `<?pi ?><a/>` - all consume the whole input, and the CDATA one
still delimits correctly right after the skipped PI. Truncation right after a
skipped PI still raises (`<?pi x?><a`, `<?pi x?><![CDATA[z`, `<?pi x?`).

Re-ran everything on the new base: FPC i386-win32 `577 assertions / 0 failed`,
FPC aarch64-linux on device `543 / 0` - the 34 delta is the guard-page
assertions, which are Windows-only. The 7 mutants are all still caught; for the
record, the two that cover the code you just touched fail as 4 assertions
(PI end test) and as an access violation (PI trim lower bound).

No rush on this - just keeping the branch directly mergeable.

PR #515, opened 2026-08-03
core: replace the unsupported GetHeapStatus with GetMemoryManagerState

vennexTSE:
### What

`RetrieveMemoryManagerInfo` (Delphi branch) reads the heap through `GetHeapStatus`. This switches it to `GetMemoryManagerState`.

### Why

`System.pas` (Delphi 13.1, compiler 37.0, build 37.0.59082.6021) declares the old API as unsupported:

```pascal
function GetHeapStatus: THeapStatus; platform; deprecated; {Unsupported}

{Returns information about the current state of the memory manager}
procedure GetMemoryManagerState(var AMemoryManagerState: TMemoryManagerState); platform;
```

The existing code is already written defensively around that (`if TotalAddrSpace <> 0 then … else result := ''`), which suggests the record has been coming back empty — with the FastMM-derived RTL manager there is nothing behind `GetHeapStatus` to fill it. If so, the Delphi heap line has silently been absent from `TSynLog` headers rather than wrong. **I could not confirm that at runtime** (the machine this was prepared on blocks launching freshly built executables), so treat it as the likely reading of the guard, not a measurement.

`GetMemoryManagerState` is the supported replacement and returns real numbers.

### Behaviour change

The reported fields change, because the two records expose different things. Small blocks are only available per size class, so they are summed over `SmallBlockTypeStates`:

```
 - Heap: Allocated=<n> Reserved=<n> Small=<n> Medium=<n> Large=<n> Blocks=<n>
```

instead of the old `AddrSpace/Uncommitted/Committed/Allocated/Free/FreeSmall/FreeBig/Unused/Overheap`. The guard becomes `if reserved <> 0`, so a replacement memory manager that reports nothing still yields `''` and no heap line, as before.

Sums are accumulated in `QWord`, so the small-block `count × size` products cannot overflow on Win32.

### Note

`GetMemoryManagerState` is itself marked `platform`, so this removes the `W1000` deprecation but not the `W1002` platform hint — there is no non-platform RTL API for memory-manager introspection. The FPC branch (`GetFPCHeapStatus`) is untouched.

### Verified

- Compiles clean on Delphi 13.1 Win32 + Win64.
- **Not** run: the `mormot2tests` regression suite, and the emitted line has not been eyeballed in a real log for the same reason as above. If you would rather see the numbers first, I can hand the diff to someone who can run it, or drop this PR and keep the other two.

---

### DIFF
--- a/src/core/mormot.core.log.pas
+++ b/src/core/mormot.core.log.pas
@@ -4199,9 +4199,42 @@ begin
 end;
 {$else}
 function RetrieveMemoryManagerInfo: RawUtf8;
+{$if defined(OSWINDOWS) and defined(ISDELPHI2007ANDUP)}
+var
+  i: PtrInt;
+  small, alloc, reserved, blocks: QWord;
+  state: TMemoryManagerState;
+{$ifend}
 begin
+  result := '';
   {$ifdef OSWINDOWS}
-  // standard Delphi memory manager
+  {$ifdef ISDELPHI2007ANDUP}
+  // standard Delphi memory manager - note: don't use the older GetHeapStatus,
+  // which the RTL declares as deprecated and unsupported
+  GetMemoryManagerState(state);
+  small := 0;
+  blocks := QWord(state.AllocatedMediumBlockCount) +
+            QWord(state.AllocatedLargeBlockCount);
+  reserved := QWord(state.ReservedMediumBlockAddressSpace) +
+              QWord(state.ReservedLargeBlockAddressSpace);
+  for i := 0 to high(state.SmallBlockTypeStates) do
+    with state.SmallBlockTypeStates[i] do
+    begin
+      inc(small, QWord(AllocatedBlockCount) * UseableBlockSize);
+      inc(reserved, ReservedAddressSpace);
+      inc(blocks, AllocatedBlockCount);
+    end;
+  alloc := small + QWord(state.TotalAllocatedMediumBlockSize) +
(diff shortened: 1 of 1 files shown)


src/core/mormot.core.log.pas
```diff
@@ -4199,21 +4199,39 @@ function RetrieveMemoryManagerInfo: RawUtf8;
 end;
 {$else}
 function RetrieveMemoryManagerInfo: RawUtf8;
+{$ifdef OSWINDOWS}
+var
```
synopse, line 4203: this would break Delphi 7 compatibility.
TMemoryManagerState was introduced with Delphi 2007 AFAIR.

synopse (changes requested):

src/core/mormot.core.log.pas
vennexTSE: You are right - fixed by guarding on `ISDELPHI2007ANDUP`, which is exactly the Delphi 2007 cutoff you named. Pre-2007 compilers keep the previous `GetHeapStatus` code unchanged, where that function is not flagged as deprecated anyway, so they stay warning-free too.

I compiled the fallback branch as well (by forcing the `{$else}` path on 13.1) so it is not just dead text: it builds, emitting only the two expected `GetHeapStatus` diagnostics that older compilers would not raise.

Also rebased on current master.

synopse (approved):
I guess the `state: TMemoryManagerState;` will make Delphi 7 fail to compiler.

But I will fix it myself during testing.

PR #517, opened 2026-08-03
TRestHttpServer with THttpApiServer engine is creating 32 threads, ignoring passed value

lzva:
TRestHttpServer was not passing aThreadPoolCount to THttpApiServer.Create


synopse (approved):
oups...

PR #520, opened 2026-08-04
core: make the USEPACKAGES conditional functional

vennexTSE:
### What

Makes the `USEPACKAGES` conditional do what it documents, as suggested in review.

Previously it documented itself as the fix for `E2201 Need imported data reference ($G) to access 'VarCopyProc'`, but nothing had read it since `77645ca93` (2020-03-29) removed its only reader (`{$ifdef USEPACKAGES}{$undef DOPATCHTRTL}{$endif}`) — so defining it could not have any effect. This PR originally proposed deleting the block; per review it now emits the switch instead.

### Change

`mormot.defines.inc`, in the Delphi *global code generation* section next to `{$A+}`:

```pascal
  {$ifdef USEPACKAGES}
    {$IMPORTEDDATA ON} // $G+ is needed when compiled within a Delphi package
  {$endif USEPACKAGES}
```

Placed inside the Delphi branch, so FPC never sees the directive. Defining `USEPACKAGES` now applies `$G+` to every mORMot unit at once, instead of relying on the package-level IDE option.

### Verified

Delphi 13.1 (compiler 37.0, build 37.0.59082.6021):

- `test/mormot2tests.dpr` full build, Win32 and Win64, with and without `USEPACKAGES`: 0 errors, byte-identical output (8,610,260 and 11,523,376 bytes code respectively).

### Important: on Delphi 13.1 this is inert

The conditional now genuinely emits the switch, but the switch does not affect E2201 on a current compiler. Tested against a real package boundary — `pkgdata.dcp` providing a global var, with its source and `.dcu` moved off the unit path so the `.dcp` is the only provider:

| access pattern | `$G-` | `$G+` | no directive | `{$IMPORTEDDATA ON}` |
| --- | --- | --- | --- | --- |
| plain read / write / `@Var` / var-param / `AtomicIncrement` | clean | clean | clean | clean |
| inline asm — `mov eax, GlobalVar` | **E2201** | **E2201** | **E2201** | **E2201** |
| statically initialized global — `var P: PInteger = @GlobalVar` | **E2201** | **E2201** | **E2201** | **E2201** |

The matrix is invariant across the switch state.

The mechanism: an imported data reference is an indirection through the import table, resolved at load time. Inline asm operands and statically initialized globals need a direct link-time address, which a package boundary cannot provide — so E2201 there reports a property of the construct, not of the switch, and `$G+` cannot help. Plain Pascal access on 37.0 always emits the indirection regardless of `$G`, so it can no longer fail.

This is consistent with the original comment's `VarCopyProc`: `VarCopyProc(Dest^, Source^)` in `mormot.core.rtti` is plain Pascal access, which older compilers did fail under `$G-` and which `$G+` did fix. Only Delphi 13.1 is installed here, so the older behaviour is not something I can verify.

The in-tree comment states this rather than implying the switch avoids the error:

```pascal
// - it will enable {$IMPORTEDDATA ON} ($G+) for all our units at once, as
// older Delphi compilers required it to access data from another package, e.g.
// "[DCC Error] E2201 Need imported data reference ($G) to access 'VarCopyProc'"
// - note: $G+ is the compiler default, and on Delphi 13.1 (compiler 37.0) this
// error is not affected by the switch any more, so defining this conditional
// has no observable effect on such a recent compiler
```

So this is an improvement over a comment prescribing a remedy with no reader, but it is not a fix for anything on a recent compiler. If you know which Delphi versions actually needed `$G+` here, that would settle whether the knob earns its place — dropping it, or bounding the note to a version range, are both reasonable and I am happy to reshape once more.

For context on the original motivation: we hit the misleading comment while configuring a build that compiles mORMot inside a BPL, enabled the conditional expecting it to act, and had to work out that it had no reader.

### DIFF
--- a/src/mormot.defines.inc
+++ b/src/mormot.defines.inc
@@ -14,8 +14,12 @@
 
 {.$define USEPACKAGES}
 // define this if you compile the unit within a Delphi package
-// - it will avoid error like "[DCC Error] E2201 Need imported data reference
-// ($G) to access 'VarCopyProc'"
+// - it will enable {$IMPORTEDDATA ON} ($G+) for all our units at once, as
+// older Delphi compilers required it to access data from another package, e.g.
+// "[DCC Error] E2201 Need imported data reference ($G) to access 'VarCopyProc'"
+// - note: $G+ is the compiler default, and on Delphi 13.1 (compiler 37.0) this
+// error is not affected by the switch any more, so defining this conditional
+// has no observable effect on such a recent compiler
 // - shall be set at the package options level, and left untouched by default
 // - note: you should probably also set "Generate DCUs only" in Project Options
 // -> Delphi Compiler -> Output C/C++ -> C/C++ output file generation
@@ -534,6 +538,10 @@
   // -- global code generation conditionals
 
   {$A+} // force normal alignment, as expected by our units
+
+  {$ifdef USEPACKAGES}
+    {$IMPORTEDDATA ON} // $G+ is needed when compiled within a Delphi package
+  {$endif USEPACKAGES}
 
   {$ifdef UNICODE}
     {$define HASVARUSTRING}


synopse:
Perhaps making it functional with the conditional {$ifdef USEPACKAGES}{$IMPORTEDDATA ON}{$endif} in mormot.defines.inc should be done.

It would document the problem and allow a project-wide fix.

vennexTSE:
Done — the PR now does exactly that instead of removing the block. Force-pushed; title and description updated to match.

`mormot.defines.inc`, in the Delphi *global code generation* section next to `{$A+}`:

```pascal
  {$ifdef USEPACKAGES}
    {$IMPORTEDDATA ON} // $G+ is needed when compiled within a Delphi package
  {$endif USEPACKAGES}
```

Inside the Delphi branch, so FPC never sees the directive. The comment now says what the conditional does.

`test/mormot2tests.dpr` builds clean on Delphi 13.1 Win32 and Win64 both with and without the conditional, with byte-identical output, so it is harmless when off.

One thing I should report rather than quietly leave out: **I could not reproduce E2201 on 13.1**, so I cannot demonstrate that this fixes the error it names — only that the switch is now genuinely emitted.

`-$G-` is honored (`G+ Use imported data references`, default ON), and I tested it across a real package boundary — a `pkgdata.dcp` providing a global var, with the unit source and `.dcu` moved off the unit path so the `.dcp` was the only provider. No E2201 from a program consuming that package, nor from a second package requiring it, with `$G-` given on the command line or written as `{$G-}` in the referencing unit, accessing the variable by value and by `@`. All clean, and `$G-` output was byte-identical to `$G+`.

So on 37.0 the switch is recorded in the DCU but does not suppress imported-data reference generation. It reads like a legacy error from compilers where `$G-` was actually enforced — which would still cover the Delphi 7+ range mORMot supports, though I have no old compiler here to confirm. You know that history far better than I can reconstruct it.

That does mean defining `USEPACKAGES` on a current compiler is in practice a no-op. It is still an improvement over a comment prescribing a remedy with no reader, but if you would prefer this be documentation only — pointing at the package `$G` option and at `NOPATCHRTL`/`NOPATCHVMT` as the successors to the `DOPATCHTRTL` that the conditional used to gate — say so and I will reshape it again.

vennexTSE:
Correction to my previous comment — I was wrong on both counts. E2201 **is** reproducible on 13.1, and `$G` turns out to be irrelevant to it. Pushed an amend with the comment wording toned down accordingly.

My earlier attempt failed because ordinary Pascal access was the only pattern I tried. Testing a range of access patterns against a real package boundary (`pkgdata.dcp` providing a global var, its source and `.dcu` off the unit path):

| access pattern | `$G-` | `$G+` | no directive | `{$IMPORTEDDATA ON}` |
| --- | --- | --- | --- | --- |
| plain read / write / `@Var` / var-param / `AtomicIncrement` | clean | clean | clean | clean |
| inline asm — `mov eax, GlobalVar` | **E2201** | **E2201** | **E2201** | **E2201** |
| statically initialized global — `var P: PInteger = @GlobalVar` | **E2201** | **E2201** | **E2201** | **E2201** |

The matrix is invariant across the switch: on compiler 37.0 the `$G` state makes no difference to E2201 in either direction.

The mechanism explains why. An imported data reference is an indirection through the import table resolved at load time. Inline asm operands and statically initialized globals need a direct link-time address, which a package boundary cannot provide — so there E2201 reports a property of the construct, not of the switch, and enabling `$G+` cannot help. Conversely, plain Pascal access on 37.0 always emits the indirection regardless of `$G`, so it can no longer fail.

That is also consistent with the original comment's `VarCopyProc`: `VarCopyProc(Dest^, Source^)` in `mormot.core.rtti` is plain Pascal access, which older compilers did fail under `$G-` and which `$G+` did fix. On 37.0 that path simply cannot fail any more.

So the honest status of this PR: the conditional now genuinely emits the switch, which is right for the older compilers that needed it, but it is **inert on current Delphi** — `test/mormot2tests.dpr` builds byte-identical with and without it on Win32 and Win64. I have only 13.1 installed, so I cannot verify the older behaviour myself.

The comment in `mormot.defines.inc` now says this outright rather than implying the switch avoids E2201:

```pascal
{.$define USEPACKAGES}
// define this if you compile the unit within a Delphi package
// - it will enable {$IMPORTEDDATA ON} ($G+) for all our units at once, as
// older Delphi compilers required it to access data from another package, e.g.
// "[DCC Error] E2201 Need imported data reference ($G) to access 'VarCopyProc'"
// - note: $G+ is the compiler default, and on Delphi 13.1 (compiler 37.0) this
// error is not affected by the switch any more, so defining this conditional
// has no observable effect on such a recent compiler
```

If you know which Delphi versions actually needed `$G+` here, that would settle whether the knob earns its place at all — you may well prefer to drop it after all, or to bound the note to a version range. Either way I am happy to reshape this once more; I would rather it be accurate than merely present.

synopse (approved):
I guess we could include it as such.
Users will give feedback about the particular version needed. I remember it was needed on some XE## versions  more or less...

PR #522, opened 2026-08-05
core: fix pending medium frees during shutdown

Moonbot-Tech:
FreeMediumPool drains Info.LastFree through _FreeMem during shutdown.

A pending small-pool backing block has the same flag value as IsLargeBlockFlag, so _FreeMem may send it to FreeLargeBlock and crash with EAccessViolation.

At this point all medium pools are about to be released anyway. The fix detaches the pending list instead of freeing those blocks individually. Leak-reporting builds still mark them free before the final pool scan.

The affected queue, flag dispatch, and pool teardown are shared by Linux and Win64. Linux is only where the failure was reproduced.

Before: deterministic exit 217 in FreeLargeBlock.
After: clean exit.

Full mORMot suite: 196,813,598 assertions, zero failures.

### DIFF
--- a/src/core/mormot.core.fpcx64mm.pas
+++ b/src/core/mormot.core.fpcx64mm.pas
@@ -3600,15 +3600,26 @@ var
   medium, nextmedium: PMediumBlockPoolHeader;
   bin: PMediumFreeBlock;
   i: PtrInt;
+  {$ifdef FPCMM_REPORTMEMORYLEAKS}
   list, next: PPointer;
+  {$endif FPCMM_REPORTMEMORYLEAKS}
 begin
+  {$ifdef FPCMM_REPORTMEMORYLEAKS}
   list := Info.LastFree;
+  {$endif FPCMM_REPORTMEMORYLEAKS}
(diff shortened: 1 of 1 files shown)


synopse:
I wonder, why is it happening only on Linux?

Moonbot-Tech:
> I wonder, why is it happening only on Linux?

I guarded the change with LINUX because Linux is the only platform where we reproduced and validated the failure.
The original failure was rare: the full memory test printed MEMORY_MEGA_PASS, then occasionally crashed in FreeLargeBlock during finalization.
We made it deterministic by allocating 800-byte blocks until a second small pool was created, freeing all but the last block from the first pool, holding its owning medium lock, and freeing the last block from another thread. This leaves the small-pool backing medium block in Info.LastFree. With the old code, finalization then consistently enters FreeLargeBlock and exits with 217.

However, I see nothing Linux-specific in the actual causal path: Info.LastFree, the header flags, the generic _FreeMem dispatch, and the pool teardown logic are shared. So I believe the Win64 path is affected as well; we simply have not reproduced and validated it there yet.

PR #524, opened 2026-08-05
core: advance to the next per-thread tiny arena

Moonbot-Tech:
When a thread's preferred tiny arena is busy, the retry calculation rebuilds the same arena address on every iteration.

As a result, the allocator tries the same arena 31 times and then falls back to a larger size class instead of checking the other arenas.

The one-instruction fix advances to the next arena while keeping the existing wrap-around mask.

Forced collision test:

- before: requested class capacity 136, returned capacity 152
- after: requested and returned capacity 136
- median allocation time: about 747 ns -> 41 ns

Full mORMot suite: 197,316,725 assertions, zero failures.

### DIFF
--- a/src/core/mormot.core.fpcx64mm.pas
+++ b/src/core/mormot.core.fpcx64mm.pas
@@ -1659,7 +1659,7 @@ asm
         sub     rbx, r8
         sub     rbx, rcx
         jz      @Sml    // from Small[rcx] to Tiny[0][rcx]
-        lea     rax, [rax + rbx - TSmallBlockInfo.Tiny] // Tiny[+1][rcx]
+        lea     rax, [rax * 2 + rbx - TSmallBlockInfo.Tiny] // Tiny[+1][rcx]
 @Sml:   {$else}
         // fair distribution among calls to reduce thread contention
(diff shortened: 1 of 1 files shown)


synopse:
Are you sure there is no overflow by selecting the next arena? It is safe because we have a + 2 everywhere for arenas?

Moonbot-Tech:
> Are you sure there is no overflow by selecting the next arena? It is safe because we have a + 2 everywhere for arenas?

Yes. There is no out-of-bounds access, and the two extra Small entries are unrelated.
Before the changed lea, the two sub instructions have converted rbx from a pointer into an arena-relative offset. The lea calculates the next arena offset only; it does not access memory.
The existing and then wraps that offset across the complete arena range before the following lea turns it back into a pointer. Therefore the sequence is:
```
Small[]     -> Tiny[0]
Tiny[i]     -> Tiny[i + 1]
Tiny[High]  -> Small[]
```
`Tiny[High + 1] `is never addressed. The retry counter also permits exactly one attempt per arena.
I checked this for all current configurations: 8 arenas in default/BOOST and 128 in BOOSTER. The two extra Small entries are only used by the later fallback to the next two size classes.

PR #527, opened 2026-08-07
core: add arenas for user-medium allocations

Moonbot-Tech:
FPCMM_BOOSTER already distributes tiny and small allocation traffic, but user-medium allocations still share one TMediumBlockInfo and its lock.

This adds FPCMM_MEDIUMPERTHREAD for Linux and Win64. Medium pools are assigned to independent arenas using the existing thread hash. Allocation probes another arena before waiting, while free and realloc recover the correct owner from the aligned pool header.

The exact arena count is workload-dependent and is not the main point of this change. This version uses four arenas:

    NumMediumBlockArenasPO2 = 2;

In our deterministic four-worker application replay, the single-arena baseline took 245.12 ns/action and four arenas took 156.96 ns/action (1.56x faster). Repeating the same source with 8 and 32 arenas gave 149.04 and 152.55 ns/action; the runs overlap, so our workload did not show a stable gain beyond four.

A deliberately sustained 32-thread medium-only benchmark continued to scale with more arenas, as expected. This confirms that the exact count is a workload policy choice rather than part of the mechanism. It remains isolated in one constant.

The exact four-arena source passed:
- 51,200 cross-thread allocate/realloc/free ownership transitions;
- repeated retained-block tests at 4 and 32 workers with identical digests;
- three deterministic application-profile processes with identical digests.

### DIFF
--- a/src/core/mormot.core.fpcx64mm.pas
+++ b/src/core/mormot.core.fpcx64mm.pas
@@ -71,7 +71,8 @@ unit mormot.core.fpcx64mm;
 // target high-end CPU/process when FPCMM_SERVER/FPCMM_BOOST are not enough
 // - will use 128 arenas for <= 256B blocks to scale on high number of cores;
 // - enable FPCMM_MULTIPLESMALLNOTWITHMEDIUM to reduce small pools locks;
-// - enable FPCMM_TINYPERTHREAD to assign threads to the 128 arenas.
+// - enable FPCMM_TINYPERTHREAD to assign threads to the 128 arenas;
+// - enable FPCMM_MEDIUMPERTHREAD for 4 user-medium arenas on Linux/Win64.
 {.$define FPCMM_BOOSTER}
 
 
@@ -133,6 +134,11 @@ unit mormot.core.fpcx64mm;
 // - warning: EXPERIMENTAL Linux and Win64 ONLY, due to very low-level asm trick
 {.$define FPCMM_TINYPERTHREAD}
 
+// use the current thread id to identify one of 4 user-medium arenas
+// - warning: EXPERIMENTAL Linux and Win64 ONLY, due to aligned OS allocations
+// and very low-level asm tricks
+{.$define FPCMM_MEDIUMPERTHREAD}
+
 // use "rep movsb/stosd" ERMS for blocks > 256 bytes instead of SSE2 "movaps"
 // - ERMS is available since Ivy Bridge, and we use "movaps" for smallest blocks
 // (to not slow down older CPUs), so it is safe to enable this on FPCMM_SERVER
@@ -182,6 +188,12 @@ interface
   {$define FPCMM_BOOST}
   {$define FPCMM_MULTIPLESMALLNOTWITHMEDIUM}
   {$define FPCMM_TINYPERTHREAD}
+  {$ifdef MSWINDOWS}
+    {$define FPCMM_MEDIUMPERTHREAD}
+  {$endif MSWINDOWS}
+  {$ifdef LINUX}
+    {$define FPCMM_MEDIUMPERTHREAD}
+  {$endif LINUX}
 {$endif FPCMM_BOOSTER}
 {$ifdef FPCMM_BOOST}
   {$define FPCMM_SERVER}
@@ -383,6 +395,7 @@ const
     {$ifdef FPCMM_SMALLNOTWITHMEDIUM}+ ' smallpool'
       {$ifdef FPCMM_MULTIPLESMALLNOTWITHMEDIUM} + 's' {$endif} {$endif}
     {$ifdef FPCMM_TINYPERTHREAD}     + ' perthrd'  {$endif}
+    {$ifdef FPCMM_MEDIUMPERTHREAD}   + ' medperthrd' {$endif}
     {$ifdef FPCMM_ERMS}              + ' erms'        {$endif}
     {$ifdef FPCMM_DEBUG}             + ' debug'       {$endif}
     {$ifdef FPCMM_REPORTMEMORYLEAKS} + ' repmemleak'  {$endif};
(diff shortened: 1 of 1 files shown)


synopse:
This is impressive work.
I wanted to try this and you made it very quickly!
What is your way of working? Using Vibe coding with high-end models?

synopse:
One comment.

When I run "2.8 Multi thread process" via `./mormot2test -t multithread` I still get some Medium sleep.
```
! All tests passed successfully.
 
 Flags: BOOSTER  assumt smallpools tinpt medpt erms repmemleak
 Small:  2K/220KB  including tiny<=256B arenas=128 pools=31
 Medium: 48MB/63MB  sleep=16
 Large:  0B/0B  sleep=0
 Total Sleep: count=16
 Small Blocks since beginning: 605K/75MB (as small=44/46 tiny=884/2032)
  48=237K  64=123K  528=52K  96=35K  128=34K  160=33K  144=30K  176=17K
  80=14K  224=8K  112=4K  32=3K  256=3K  288=1K  2176=805  352=633
 Small Blocks current: 2K/220KB
  48=1K  64=440  352=139  32=105  128=85  112=64  80=39  96=28
  144=14  288=9  192=9  576=7  240=6  416=6  160=5  1376=5
```
Perhaps there is still some room to improvement.

Of course, it seems better than plain FPCMM_SERVER:
```
All tests passed successfully.
 
 Flags: SERVER  assumt erms debug repmemleak
 Small:  2K/220KB  including tiny<=128B arenas=8 fed from Medium
 Medium: 6MB/56MB    peak=11MB current=5 alloc=45 free=40 sleep=126
 Large:  0B/0B    peak=0B current=0 alloc=0 free=0 sleep=0
 Total Sleep: count=126
 Small Blocks since beginning: 604K/75MB (as small=44/46 tiny=56/56)
  48=237K  64=123K  528=52K  96=35K  128=34K  160=33K  144=29K  176=17K
  80=14K  224=8K  112=4K  32=3K  256=2K  288=1K  2176=790  352=639
 Small Blocks current: 2K/220KB
  48=1K  64=440  352=139  32=105  128=88  112=64  80=39  96=28
  144=14  192=9  288=9  576=7  240=6  416=6  1056=5  1376=5
  ```

synopse:
FYI some refactoring was done today in fpcx64mm.
Mostly register allocation and code cleaning.

I wonder if you may find something on your side.
The main change about performance was to enable `FPCMM_TINYPERTHREAD` in FPCMM_SERVER because on my side it makes a noticeable difference not in micro-benchmarks but with real-world process involving a few threads. Data locality per thread makes a noticeable difference. Nice to observe that thread affinity is a good idea. Perhaps we have the best of both world: `threadvar` like performance without the need of any GC logic.

Moonbot-Tech:
Arnaud, thank you for the `fpcx64mm` rework. We ran it separately on Win64 and
Linux and found two independent points:

1. [#546](https://github.com/synopse/mORMot2/pull/546) - on Linux the large-allocation path tested `RCX`, while the System V
   ABI passes `size` in `RDI`. We added a deterministic regression test.
2. [#547](https://github.com/synopse/mORMot2/pull/547) - `RSI` improves `_FreeMem` on System V, but produced a reproducible
   slowdown of about 6.9% in a Win64 same-class ring. The proposed version uses
   `RBX` on Win64 and keeps `RSI` on Linux.

We also tried to break the new version with a broad cross-thread stress test:
contention, ownership transfers between threads, and mixed alloc/free/realloc
sequences. We could not :) No additional correctness issue was found.

synopse:
Out of context, what is your usage of the MM unit?

Moonbot-Tech:
Funny story. A month ago we simply decided to build our trading terminal (MoonBot, Delphi/Windows) for Linux with FPC — and that attempt pulled us so deep into working on FPC itself that we almost forgot the original goal :) It started with a couple of simple bugs and inefficiencies, one thing led to another — as of today we're at roughly 200 individual repairs.

Your MM became the center of the whole idea: "compiler + fpcx64mm" as the most efficient way to build Delphi projects for Linux. It serves us twice — as the production MM under real trading load (many worker threads, bursty allocations on the network hot path — that's where the arena PR came from), and as the bundled default MM of the compiler itself.

By the way, we're getting close to finishing what we set out to do. What do you think — would the community find it interesting if we published it on GitHub? The ideology: an FPC fork aiming at oracle-verified Delphi 12 compatibility (semantics checked against DCC64 by a large regression battery and a structural fuzzer); Win64 + Linux x86-64 only; `string` = UTF-16 as in Delphi on both OSes, with UTF-8 exactly at the POSIX boundary; fpcx64mm as the default MM; Delphi RTTI and the full managed-record lifecycle (Initialize/Finalize/Assign); exact 128-bit Currency arithmetic. It's not a "better FPC" — it's a "Delphi target for Linux"; vanilla stays vanilla.

Also by the way: for the compiler's qualification we run the full mORMot test suite (several hundred million assertions per run) — and for that we apply a small Unicode patch to it (FPC + UTF-16 string on POSIX: transcoding at the raw API boundaries, with every branch dead for the existing combinations — for Delphi and regular FPC the code stays byte-for-byte yours). Do you think that would be interesting as an upstream PR — or is FPC/Unicode out of your scope and it should just live on our side as a patch?

synopse:
Very interesting!
Perhaps the full Unicode path could be good. But I wonder what is missing about Delphi for POSIX support in our code base with Unicode. You mean UNICODEMODE in FPC and TFileName = string = UnicodeString everywhere in mORMot? I guess we could see if we could integrate it.

I have made a small analysis against the libc allocator.
We are in the right direction for sure.
https://synopse.info/forum/viewtopic.php?id=7597

Moonbot-Tech:
Arnaud, a short follow-up to our discussion above. We have finished and published the project we mentioned:

**[MoonCompiler 1.0.0](https://github.com/Moonbot-Tech/MoonCompiler)** — [release](https://github.com/Moonbot-Tech/MoonCompiler/releases/tag/v1.0.0) · [MM documentation](https://github.com/Moonbot-Tech/MoonCompiler/blob/main/doc/MEMORY_MANAGER.md).

Thank you for `fpcx64mm`: it ended up becoming one of the central parts of the whole system. We use it as the product memory manager, and qualify and benchmark the compiler, RTL, and MM together. In the final Pulse benchmark run, with the same compiler and workloads, the allocator group using `fpcx64mm` ran on average 1.65 times faster than with the standard FPC MM.

Please take a look. We would really value your opinion—both on the `fpcx64mm` integration and on MoonCompiler as a whole.

synopse:
Very nice!

Perhaps FPCMM_MOONSHARD could be also included as another option in the mainstream MM.

PR #530, opened 2026-08-08
core: fix XmlToVariant() xpoVariantGuessType corrupting text values

zen010101:
Following up on the #487 discussion about `xpoVariantGuessType` ("may be used only for some very proven data input"): even proven input can contain dotted identifiers or long decimal IDs, and the relaxed number parsing then corrupts them silently:

| input | guessed as |
|---|---|
| `1.2.3` | `0.123` |
| `10.0.0.1` | `0.10001` |
| `9223372036854775808` | `-9223372036854775808` (Int64 wrap) |
| `-9223372036854775809` | `9223372036854775807` (sign flip) |
| `34020000001320000001` (GB28181 SIP ID) | `3.4020000001320002E19` (lossy double) |

This PR adds a small `CanGuessType()` gate before `GetVariantFromNotStringJson()` in `TXmlParser.ToDocVariant`, reusing existing primitives only (`IsInt64` + `ToInt64` + `IsConstantJson` + `IsNumberJson`): only exact JSON literals are converted, and integer-looking text out of Int64 range remains lossless text rather than an approximate double - a 20-digit decimal is an identifier, not a measurement. `GetVariantFromNotStringJson()` itself is untouched, so JSON parsing is unaffected.

Two commits: the first only adds tests locking the agreed-safe envelope (green on current master), the second is the fix with 6 more tests - red before it on exactly the 5 corruptions above, green after. One deliberate consequence, documented in the interface comment: text with leading/trailing blanks is not guessed any more, as a side effect of the strict JSON lexing.

Verified on i386-win32 (FPC 3.2.3): whole `TTestCoreProcess` 0 / 14,929,618 failed, based on current master (`95127a13d`).

### DIFF
--- a/src/core/mormot.core.fmt.pas
+++ b/src/core/mormot.core.fmt.pas
@@ -245,7 +245,10 @@ type
   // which are silently skipped by default
   // - xpoKeepWhiteSpace would return xtText tokens made only of whitespace,
   // which are silently skipped by default
-  // - xpoVariantGuessType let XmlToVariant() recognize booleans and numbers
+  // - xpoVariantGuessType let XmlToVariant() recognize booleans and numbers,
+  // matching only exact JSON literals: anything else is kept as text, e.g.
+  // '1.2.3' '192.168.0.1' '007' or integers out of Int64 range (a 20-digit
+  // decimal identifier should not become an approximate double)
   TXmlParserOption = (
     xpoNoException,
     xpoStripNamespacePrefix,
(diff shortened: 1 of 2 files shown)


synopse:
This is clearly not the right path.
I guess GetVariantFromNotStringJson() should be fixed to be more strict about the conversion.

zen010101:
Understood, and no problem closing this.

For our own use (GB28181 / ONVIF) we don't rely on guessing at all - the `GetU` / `GetI` / `ConsumeText` path you added is exactly what we needed, since the caller knows the target type. So this is not a blocker for us either way; whether `GetVariantFromNotStringJson()` should be tightened is entirely your call. Thanks for the quick look.

PR #533, opened 2026-08-10
core: JsonToXml() and new VariantToXml() support for the '@attribute' and '#text' conventions

tobfel:
### Problem

`XmlToVariant()` / `XmlToJson()` map XML attributes to `@name` fields and mixed content to a `#text` field. The writing direction did not know about those conventions and emitted them as element names, i.e. `<@name>` and `<#text>`, which are not well-formed XML — so `XmlToJson()` output could not be turned back into XML, and attribute-based formats (SOAP, EPC, payment terminals, ...) had to be assembled by hand with `XmlEscape()`.

```pascal
u := '<a><b>1</b><b>2</b><c d="x">t</c></a>';
JsonToXml(XmlToJson(u), '', '', JXO_XML)  // = u
```

### Change

Two ways to write those conventions back out:

```pascal
type
  TJsonToXmlOption = (jxoAttribute, jxoText);
  TJsonToXmlOptions = set of TJsonToXmlOption;

const
  JXO_XML = [jxoAttribute, jxoText];

// JSON buffer based - options default to [] i.e. the previous behaviour
function AddJsonToXml(W: TTextWriter; Json: PUtf8Char; ArrayName: PUtf8Char = nil;
  EndOfObject: PUtf8Char = nil; Options: TJsonToXmlOptions = []): PUtf8Char;
procedure JsonBufferToXml(P: PUtf8Char; const Header, NameSpace: RawUtf8;
  out result: RawUtf8; Options: TJsonToXmlOptions = []);
function JsonToXml(const Json: RawUtf8; const Header: RawUtf8 = XMLUTF8_HEADER;
  const NameSpace: RawUtf8 = ''; Options: TJsonToXmlOptions = []): RawUtf8;

// TDocVariantData DOM based - the reverse of XmlToVariant()
procedure AddVariantToXml(W: TTextWriter; const Doc: variant;
  Options: TJsonToXmlOptions = JXO_XML);
function VariantToXml(const Doc: variant; const Header: RawUtf8 = XMLUTF8_HEADER;
  const NameSpace: RawUtf8 = ''; Options: TJsonToXmlOptions = JXO_XML): RawUtf8;
```

The mapping is **opt-in**: with the default `[]` every existing `JsonToXml()` caller keeps its exact previous output. All signatures stay source-compatible through defaulted parameters.

### How

`AddJsonToXml()` is a single destructive pass (`GetJsonField()` unescapes in place), and the parent used to write the whole start tag before it could know whether the child begins with attributes. The `>` is now deferred: the parent emits `<name` only, and the recursive call closes the tag after the leading `@name` fields, or before the first content. Per review, the recursion state moved into a single `var` context:

```pascal
TAddJsonToXmlNodeContext = record
  W: TTextWriter;
  Json: PUtf8Char; // updated in place - nil on any error
  Options: TJsonToXmlOptions;
end;

procedure AddJsonToXmlNode(var Context: TAddJsonToXmlNodeContext;
  ArrayName, EndOfObject: PUtf8Char; Pending: boolean);
```

so no level threads the buffer position through its result, and `AddJsonToXml()` stays a thin public wrapper. Array items are passed `Pending=true` too, so they may carry attributes. `AddXmlEscape()` is reused unchanged for attribute values — it already escapes `<`, `>`, `&`, `"`, `'` and TAB/LF/CR.

`AddVariantToXml()` walks `Names[]` / `Values[]` directly. Because the whole object is available at once, the attributes are collected in a first pass, so — unlike the JSON path — a `@name` field may appear **anywhere** in its object:

```pascal
VariantToXml(_Json('{"a":{"b":"1","@d":"x"}}'), '') = '<a d="x"><b>1</b></a>'
```

The JSON-buffer path keeps the documented restriction that `@name` fields come first, which is how `XmlToVariant()` generates them. A `@name` field that cannot be represented as an attribute (appearing after content, or holding an object/array) is skipped. An attribute-only object yields `<a x="1"></a>`; a self-closing form would need a flag back to the caller to suppress the closing tag, so it is left out.

The name space closing-tag logic is now shared by `JsonBufferToXml()` and `VariantToXml()` as `AddXmlNameSpaceEnd()`.

### Tests

`TTestCoreProcess.XmlToVariant` in `test/test.core.data.pas` extended next to the existing `XmlToJson()` checks, covering both directions, the escaping, the opt-out default, and `XmlToVariant()`/`XmlToJson()` round-trips.

Verified on Win32 with **Delphi 13.1** and **Delphi 7** (D7-compatible: no generics, no inline variables):

* full regression suite green — the only red check on this machine is the pre-existing `Dmi smbios` OEM-string test in *Core base*, which fails identically on unmodified master here;
* `Core process` 0 failed / 15,068,707 assertions, XML sub-suite 3,794 assertions;
* a separate 44-case probe on both compilers covering the edge cases: `{"a":{"#":"x"}}` and `{"a":{"#texts":"x"}}` stay plain elements, `{"a":{"@":"x"}}` and non-scalar `@` values are skipped, `XmlToJson(JsonToXml(...))` reparses to the original JSON.

Design and wording were reviewed and approved in tobfel/mORMot2#1.


tobfel:
@synopse - i have created a new pr that also support "shortform-/selfclosing" Tags (https://github.com/tobfel/mORMot2/pull/2)

Issue #532, opened 2026-08-08
JsonToXml() does not support the '@attribute' / '#text' conventions used by XmlToVariant()


tobfel:
> **Working implementation:** tobfel/mORMot2#1 (draft) —
> branch [`feature/jsontoxml-attributes`](https://github.com/tobfel/mORMot2/tree/feature/jsontoxml-attributes),
> one commit on top of current master. Happy to re-target it as a PR here if the
> direction suits you.

### Summary

`XmlToVariant()` / `XmlToJson()` map XML attributes to `@name` fields and mixed
content to a `#text` field. `JsonToXml()` does not know about those conventions, so
it emits `<@name>` and `<#text>` elements, which are not well-formed XML (`@` and `#`
are not valid XML name start characters). As a result the round-trip is broken, and
generating attribute-based XML (SOAP, EPC, O.P.I. payment terminals, DATEV, ...) has
to be done by hand with `XmlEscape()`.

### Reproduce

```pascal
u := '<a><b>1</b><b>2</b><c d="x">t</c></a>';
j := XmlToJson(u);
// j = '{"a":{"b":["1","2"],"c":{"@d":"x","#text":"t"}}}'   <- as checked in test.core.data
x := JsonToXml(j, '');
// x = '<a><b>1</b><b>2</b><c><@d>x</@d><#text>t</#text></c></a>'
// expected: '<a><b>1</b><b>2</b><c d="x">t</c></a>'
```

The current output is rejected by every XML parser, including mORMot's own
`TXmlParser`, so no existing code can depend on it.

### Proposal

Teach `AddJsonToXml()` the same conventions in the writing direction:

* a field named `@xxx` holding a scalar becomes an attribute `xxx="..."` of the
  enclosing element;
* a field named `#text` becomes the element's text content;
* everything else is unchanged.

Because the current output is invalid XML, I would make this the default and add an
opt-out rather than an opt-in:

```pascal
type
  TJsonToXmlOption = (jxoNoAttribute, jxoNoText);
  TJsonToXmlOptions = set of TJsonToXmlOption;

function AddJsonToXml(W: TTextWriter; Json: PUtf8Char; ArrayName: PUtf8Char = nil;
  EndOfObject: PUtf8Char = nil; Options: TJsonToXmlOptions = []): PUtf8Char;
procedure JsonBufferToXml(P: PUtf8Char; const Header, NameSpace: RawUtf8;
  out result: RawUtf8; Options: TJsonToXmlOptions = []);
[...]

PR #535, opened 2026-08-10
core: add jxoSelfClosed to write void XML elements in short form

tobfel:
Follow-up optimization on top of #532 / #533, rebuilt on your e4fa0efef refactoring.

> **Stacked on #534.** The first commit of this branch is the `VariantToXml()`
> void-document fix from that PR, because this change touches the same code and
> needs it (`{"a":{}}` has to become `<a/>`, not `<a>{}</a>`). Merging #534 first
> makes it disappear from this diff.

## What

An element with no text and no sub-element is always written in the long form:

```
{"a":{"@d":"x"}}  ->  <a d="x"></a>
{"a":""}          ->  <a></a>
{"a":{}}          ->  <a></a>
```

`jxoSelfClosed` writes the self-closing short form instead:

```
{"a":{"@d":"x"}}  ->  <a d="x"/>
{"a":""}          ->  <a/>
{"a":{}}          ->  <a/>
```

Opt-in like `jxoAttribute`/`jxoText`, and deliberately **not** part of
`JXO_ENABLED`, so `[]` and `JXO_ENABLED` output stay byte-for-byte as before.

## Why

Without it, a caller who wants the short form has to fall back to hand-made
string concatenation for the whole document - which is what #532 set out to
remove in the first place. It is a real constraint for attribute-heavy formats,
where void elements are the common case and the long form roughly doubles the
element count on the wire.

Nothing is lost by choosing either form: `XmlToVariant()` reads `<a/>` and
`<a></a>` back as the very same content, so this is purely about the emitted
layout.

## How

Both writers need to know whether the nested level wrote any content at all:

- `AddJsonToXml()` takes `Pending` as an optional `PBoolean` instead of a
  `boolean`, so the same parameter also works as a return channel - a still set
  `Pending^` means the nested level wrote nothing, and the caller emits `/>`
  instead of `></name>`. Same arity, same default, and it stays the internal
  flag you already documented it as. The alternative would have been a second
  out-parameter next to it; this seemed the smaller surface.
- `AddVariantToXmlNode()` gets the flag as a `var` parameter (it is
  implementation-only), and the `>` of the start tag is deferred until content
  is actually written, so a `#text` field holding a void value no longer forces
  the long form. `VariantToXmlText()` was split out of `AddVariantToXmlText()`
  to test for a void value before anything is emitted.

## Tests

26 new `CheckEqual` in `XmlParser`, covering both writers: void values,
attribute-only elements, void `#text`, array items, nested levels, the
unaffected non-void cases, and that both forms are read back identically.
All previous assertions are unchanged.

Verified on **Delphi 7** and **Delphi 13 Win32** with a standalone harness
replicating the whole `XmlToJson`/`JsonToXml`/`VariantToXml` assertion set plus
the new cases: 70/70 and 48/48 on both compilers.

### DIFF
--- a/src/core/mormot.core.fmt.pas
+++ b/src/core/mormot.core.fmt.pas
@@ -1967,25 +1967,25 @@ var
   d: PDocVariantData;
 begin
   d := _Safe(Value);
-  if d^.Count <> 0 then
-  begin
-    if d^.IsArray then
-    begin // arrays are written as a list of items, without any root
-      for i := 0 to d^.Count - 1 do
-        AddVariantToXmlValue(W, Name, d^.Values[i], Options);
-      exit;
-    end;
-    W.Add('<');
-    AddXmlEscape(W, pointer(Name));
-    if jxoAttribute in Options then
+  if d^.IsArray then
+  begin // arrays are written as a list of items, without any root
+    for i := 0 to d^.Count - 1 do
+      AddVariantToXmlValue(W, Name, d^.Values[i], Options);
+    exit; // a void array writes no element at all, as AddJsonToXml() does
+  end;
+  W.Add('<');
+  AddXmlEscape(W, pointer(Name));
+  if d^.IsObject then
+  begin // a document is never written as text, even if it is void
+    if (d^.Count <> 0) and
+       (jxoAttribute in Options) then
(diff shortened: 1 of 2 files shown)


synopse (approved):
I still don't like the styling code but if it is correct, why not.

PR #536, opened 2026-08-11
core: fix TDwarfReader.ReadAbbrevTable rejecting vendor DWARF attributes

zen010101:
### Problem

Since 05967413f, `TDwarfDebugAbbrev.Attrs[].attr` is a byte, and
`ReadAbbrevTable` raises `DWARF: a=% f=% n=% overflow` for any attribute
code above 255.

gcc emits vendor attribute codes in the `$2111..$2138` range
(`DW_AT_GNU_all_call_sites`, `DW_AT_GNU_locviews`,
`DW_AT_GNU_call_site_value`, ...) as soon as C objects compiled with `-g`
are statically linked into an FPC executable — very common in practice
(any FPC program linking a C library built with debug info).

The result: `TDebugFile.Create()` aborts, its `except` clause clears
everything, `HasDebugInfo` stays `false`, and **no `.mab` is generated at
all** — silently losing stack symbolization for the *whole* program, not
just the C part.

### Fix

Such vendor attributes only need to be *skipped*, and skipping is driven by
the attribute's **form**, not its code. So when `a > 255`, store
`attr := 0` (which matches no `DW_AT_*` constant used later) and keep the
form. This preserves the byte-sized `Attrs[].attr` (no memory regression
vs. widening it) while no longer rejecting the abbrev table.

### Verification

- 5.9 MB aarch64 FPC binary statically linking 15 gcc compilation units:
  **0 symbols before -> 13183 symbols / 230 units after**.
- Reproduced on i386-win32 against a small gcc-produced DWARF fixture:
  `HasDebugInfo` flips `false -> true` (43 line entries parsed), while
  regular FPC DWARF symbolization is unchanged.

One commit on top of current master.


synopse (approved):
I indeed focused on FPC-generated DWARF.

PR #538, opened 2026-08-13
openapi: generate PascalCase enum items from ALL_CAPS values

sovdchains:
Specs declaring `enum` values in SCREAMING_SNAKE_CASE produce unreadable identifiers: `CamelCase()` removes each `_` and uppercases the letter after it, but never lowercases anything, so `USER_NOT_FOUND` becomes `urUSERNOTFOUND`.

New opt-in `opoEnumPascalCase` option: `TPascalEnum.ToTypeDefinition` uses `IsUpper()` to detect choices without any lowercase letter and lowercases them before `CamelCase()`, emitting `urUserNotFound`. Mixed-case values pass through unchanged, and the `_TEnumName` const arrays keep the original spec strings, so the JSON wire format is unaffected.


synopse:
Makes sense!

Thanks for the feedback! :)

PR #540, opened 2026-08-16
core: fix memory corruption on small-block contention

Moonbot-Tech:
## Cause

`GetmemSleepCount` has 44 entries, one per irregular unique small-block class, but both contention paths index it as `(BlockSize div 16) - 1`. For the 2608-byte class this is index 162, so the rare sleep-after-contention path writes into the following `TSmallBlockInfo` fields.

The mismatch was introduced when the former per-class counter was moved to the shared array in `0f23b9fcf9`.

## Fix

Use one `NumSmallBlockGranularitySlots` constant for both `GetmemLookup` and `GetmemSleepCount`, remove the now-unused unique-class count, and assert their shared index-domain invariant during MM initialization.

The allocator assembly and allocation hot path are unchanged. Static metadata grows by 476 bytes.

## Reproduction

With `FPCMM_SERVER` and `FPCMM_NOPAUSE`, 32 threads repeatedly execute `GetMem(2600)` / `FreeMem` and then read `CurrentHeapStatus` / `GetSmallBlockContention`.

- current master: 20/20 runs terminate with exit 217 after `small-sleeps=0 entries=0`
- fixed: 20/20 runs pass and report a non-zero counter for block size 2608

The same A/B result was reproduced on Win64 and Linux.

## Validation

- layout invariant: old source fails at initialization; fixed source passes O-, O2 and O3 on Win64
- focused contention regression: 20/20 passes on Win64 and 20/20 on Linux
- `TTestCoreBase._fpcx64mm`: 14/14 on Win64 (FPC 3.2.2) and Linux
- full Linux suite on the immediate pre-refactor parent: 202,489,279 assertions, 0 assertion failures; the existing `TTestCoreBase.Debugging` exception still leaves the process with exit 1
- current master full-suite compilation is independently blocked before tests by the unrelated missing `cvCrlFailed` identifier in today's crypt refactor

### DIFF
--- a/src/core/mormot.core.fpcx64mm.pas
+++ b/src/core/mormot.core.fpcx64mm.pas
@@ -945,7 +945,6 @@ const
   {$endif FPCMM_BOOSTER}
 
   NumSmallBlockTypes       = 46;
-  NumSmallBlockTypesUnique = NumSmallBlockTypes - 2; // last 2 are redundant
   MaximumSmallBlockSize    = 2608;
   NumTinyBlockTypes        =
      1 shl NumTinyBlockTypesPO2; // 8 (128B) or 16 (256B)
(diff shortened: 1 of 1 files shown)


synopse (approved):

synopse:
Thanks for the feedback!

Issue #292, opened 2024-08-31
Is there any future plan to support HTTP Server side File Upload?


Coldzer0:
Hello @synopse 

I'd like to know if there are any plans to support HTTP server file upload.

I have a client code that does a `THttpMultiPartStream` and sends the stream to the server.
But in my testing, there's no clear way in mormot2 to handle this kind of file upload, especially if the files are big (> 2GB).

If there are no plans to support this in the near future, Can you help by giving instructions on how to approach this in mormot2?
 `( Like how to handle the data by putting it directly into a file stream and not keeping it in memory without messing with the source code of mormot itself )`

mORMot2 code is vast, and even after a couple of years of working with it, I still find new stuff and learn from it ^_^

synopse:
It is indeed not yet supported.

But it is planned in the close future, with the upcoming proxy and reverse proxy features of our async server.
We need this, because nginx is no option on Windows, and our async server using IOCP seems to be pretty fast, so we will continue to enhance it.

So stay tuned, we will support it, I hope before the end of this year.

synopse:
Note that all the internal logic for using a TStream instead of a RawByteString is already there.
We just need a mean to trigger it properly.

Perhaps in OnBeforeBody or - perhaps better - with a route registration.
What do you think?

Coldzer0:
I was writing notes about building something using OnBeforeBody and Parsing the Headers. 
And then we know we have the `Content-Type` to detect the `multipart` file upload and the `Content-Length`.
And for sure, the `boundary` string means that we know this is the end of file data and that we have started a new file.

```Pascal
var 
  Files : THttpMultiPartStream;  
Begin
  Files := THttpMultiPartStream.Create;
  Files.AddFile('files', 'F:\UploadTest\hello.txt');
  Files.AddFile('files', 'F:\UploadTest\world.txt');
  Files.Flush; 
  ...
End;
```

Headers for it on the server side look like this 
```yaml
Cache-Control: no-cache
Connection: Close
Pragma: no-cache
Content-Type: multipart/form-data; boundary=oKhwAf29zfOYjwQy
Accept: */*
User-Agent: Mozilla/5.0 (Win x64; mORMot) WH/2 FileUpload_Debug64
Content-Length: 368
Host: localhost
```
The content will be 
```yaml
--oKhwAf29zfOYjwQy
Content-Disposition: form-data; name="files"; filename="hello.txt"
Content-Type: text/plain; charset=UTF-8
[...]

PR #541, opened 2026-08-17
New THttpServerGeneric.OnBodyDownload event to stream a request body

landrix:
Server-side counterpart of #505, and a first step for #292: let a handler
receive a huge upload without buffering it in memory.

A new `THttpServerGeneric.OnBodyDownload` event is called once the headers
are parsed (and `OnBeforeBody` accepted the request), and can return a
`TStream` - typically a `TFileStreamEx` spooling into a temporary file - to
receive the incoming body. Returning `nil` keeps the current in-memory
behavior. Implemented for both socket-based families, `THttpServer` and
`THttpAsyncServer`, so memory usage stays constant whatever the body size.

The request then reaches the callbacks with `InContentType =
STATICFILE_CONTENT_TYPE` and `InContent` = the local file name, mirroring
the response process; the file is deleted once the request has been
processed, unless the handler renamed or moved it to take ownership. The
original `Content-Type` remains available from `InHeaders`.

Error paths:
- a compressed body (`Content-Encoding:` matching a registered algorithm)
  can not be streamed: 415 + `Accept-Encoding: identity`
- a chunked body over `MaximumAllowedContentLength`: 413 while receiving,
  since its final size is not known in advance
- a failing write into the stream (e.g. a full disk): 507, as best effort -
  the client may still be sending - through the new
  `HTTP_INSUFFICIENTSTORAGE` constant and a `hrsErrorContentStreamWrite`
  state

Two pre-existing bugs showed up while testing this, fixed in the same
branch (both only reachable on `THttpAsyncServer`, and unrelated to the
event itself):
- `THttpRequestContext.ProcessRead` never consumed the chunk data bytes
  from its input buffer in `hrsGetBodyChunkedData` (neither `st.P` nor
  `st.Len`), so the following lines were parsed out of the chunk payload -
  any chunked request body stalled or was rejected
- `DoProcessParseLine` read `P[Len + 1]` without checking the buffer end,
  so a packet ending exactly on the `#13` of a CRLF caused a one byte
  out-of-bounds read, and a stray `#10` there made `st.Len` go negative

Tested by a new `TNetworkProtocols.BodyDownload` method (179 assertions,
both families): spooling and integrity, in-memory fallback, multipart
decoded straight from the spooled file with `THttpMultiPartDecoder`,
chunked bodies, exact `MaximumAllowedContentLength` boundary, 415/413/507,
aborted uploads and spool cleanup. Green with FPC on aarch64-linux and
Delphi Win32/Win64. @zen010101 also ran an extensive matrix on a Rockchip
RK3588 (aarch64, FPC 3.2.3): 111 MB and 4.5 GB uploads, concurrent, rate
limited, real disk-full, keep-alive - all byte-identical by SHA-256, with
no spool leftovers.

Still open, from your first-pass feedback: the stream lifetime. My
suggestion was a `TFileStreamEventuallyDelete = class(TFileStreamEx)` with
a `DeleteFileOnDestroy` property, which would drop `ContentInputName`
entirely - the deletion policy would live in the class, like nginx's
`client_body_in_file_only clean` vs. `on` - plus a
`THttpServerRequestAbstract.InContentStream` property handing the already
open instance to the request instead of reopening it by name.

### DIFF
--- a/src/core/mormot.core.os.pas
+++ b/src/core/mormot.core.os.pas
@@ -193,6 +193,8 @@ const
   HTTP_GATEWAYTIMEOUT = 504;
   /// HTTP Status Code for "HTTP Version Not Supported"
   HTTP_HTTPVERSIONNONSUPPORTED = 505;
+  /// HTTP Status Code for "Insufficient Storage"
+  HTTP_INSUFFICIENTSTORAGE = 507;
 
   /// a fake response code, generated for client side panic failure/exception
   // - for it is the number of a man, and that number is 666
--- a/src/core/mormot.core.text.pas
+++ b/src/core/mormot.core.text.pas
@@ -10536,7 +10536,7 @@ end;
 
 const
   // last item is fake HTTP status 513 = 'Invalid Request'
-  INDEX_HTTP_INVALID = 46;
+  INDEX_HTTP_INVALID = 47;
   // sorted by actual usage order for WordScanIndex() in matching HTTP_CODE[]
   HTTP_REASON: array[0 .. INDEX_HTTP_INVALID] of RawUtf8 = (
    'OK',                                // HTTP_SUCCESS - should be first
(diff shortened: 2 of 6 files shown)


synopse (approved):
Thanks!

PR #542, opened 2026-08-17
net: fix out-of-bounds write when decoding a chunked body in memory

landrix:
THttpRequestContext.ProcessRead sets fContentPos to @Content[length(Content)] right after SetLength(Content, bytes), i.e. past the end of the reallocated buffer instead of at the previous end. Every chunk is then written out of bounds. A plain THttpAsyncServer with no OnBodyDownload segfaults on any Transfer-Encoding: chunked request - reproduced here both at the ProcessRead level and end-to-end.

Present since c8d084e2. The in-memory chunked path has no test coverage: the chunked test from #541 always goes through a spool stream, so it takes the ContentStream <> nil branch. Added a multi-chunk request to /mem in HttpBodyDownload, which crashes the test process without the fix.

Gate: FPC aarch64/Linux TNetworkProtocols 0/134061, Delphi Win32 and Win64 0/297.


synopse (approved):
I will fix manually the merging issues.

Issue #543, opened 2026-08-17
TSynEvent.WaitFor can report a timeout on an event which was set (hits TPipeStream)


landrix:
`TSynEvent.WaitFor` can return `false` — i.e. report a timeout — on an event that was set, and it does so immediately rather than after the timeout. A waiter that treats `false` as "nothing happened" then drops a signal it actually received.

### Why it happens

`TSynEvent` keeps its state in two places, and `SetEvent` updates them in two steps with no atomicity in between (`mormot.core.os.posix.inc`):

```pascal
procedure TSynEvent.SetEvent;
begin
  fNotified := true; // should be set before notification
  if fFD <> 0 then
    LinuxEventFDWrite(fFD, 1)
  ...

procedure TSynEvent.ResetEvent;
begin
  if fFD <> 0 then
    if fNotified and
       WaitReadPending(fFD, 0) then
      LinuxEventFDRead(fFD);
  ...
  fNotified := false;

function TSynEvent.WaitFor(TimeoutMS: cardinal): boolean;
  ...
  result := fNotified;
```

Interleaving of a signalling thread S and a waiting thread W:

1. S enters `SetEvent`, sets `fNotified := true`, and is preempted before writing the eventfd.
2. W calls `ResetEvent`: `fNotified` is true, but `WaitReadPending(fFD, 0)` is false because the eventfd write has not happened yet, so nothing is drained — and `fNotified := false`.
3. S writes the eventfd: the counter is now 1, while `fNotified` is false.
4. W calls `WaitFor`: the fd is immediately readable, so it is drained, and `WaitFor` returns `fNotified` — **false**. The caller sees a timeout that did not happen.

The state stays poisoned in the sense that `ResetEvent` can only drain when `fNotified` is true, so a stale counter can survive.

The Windows implementation in `mormot.core.os.windows.inc` has the same shape: `fNotified` is set before `SetEvent(fHandle)`, `ResetEvent` clears it, and `WaitFor` returns `fNotified`.

### Reproducer

Self-contained, only `mormot.core.base` and `mormot.core.os`. Two threads per pair share a counter under a `TLightLock`; the signaller calls `SetEvent` outside the lock and the waiter calls `ResetEvent` under it — the pattern `TPipeStream` uses.

```pascal
program syneventrace;

{$I mormot.defines.inc}

uses
  {$ifdef UNIX} cthreads, {$endif}
  sysutils, classes, mormot.core.base, mormot.core.os;

type
  TShared = class
  public
[...]

PR #544, opened 2026-08-17
net: OnBodyDownload stream lifetime - TFileStreamEventuallyDelete, rfContentFileNameNeedDelete, InContentStream

landrix:
Follow-up to #541, addressing the stream lifetime question you raised in its review.

You wrote: *"TFileStreamEventuallyDelete could be the most simple solution. But a new rfContentFileNameNeedDelete item in ResponseFlags could make sense too, because it could work with any kind of stream. We could have both anyway."* — so this has both.

## What is in here

**`TFileStreamEventuallyDelete`** (mormot.core.os) — a `TFileStreamEx` deleting its own file in `Destroy`, with `DeleteFileOnDestroy` to switch it off. The spool file lifetime is then bound to the stream instance, with nothing to track.

**`rfContentFileNameNeedDelete`** (ResponseFlags) — the other half: the server owns the `ContentInputName` file. It is set for any `TFileStreamEx` which is not a `TFileStreamEventuallyDelete`, so the deletion no longer depends on which class the callback returned. Note this fills the flags byte exactly — 8 items now, and both comments in the code promise a single byte.

The two combine into the nginx pair: `client_body_in_file_only clean` is the default, and `on` is a `TFileStreamEventuallyDelete` with `DeleteFileOnDestroy := false`, which the server then leaves alone.

**`THttpServerRequestAbstract.InContentStream`** — the part that makes the rest useful. The body stream now survives until the request has been processed instead of being freed before `Prepare`, and is handed to the handler, rewinded if seekable. `THttpRequestContext.ContentInputDone` moves it out of `ContentStream` (which the response reuses) into the new `ContentInputStream`, and `ProcessDone` releases it at the end of the request. This replaces the early free in both server families.

That last point is what lets a non-file stream be used at all: before, any `TStream` other than a `TFileStreamEx` was written to and then destroyed before `OnRequest`, so the handler never saw it. Now a `TRawByteStringStream`, a pipe or any custom stream can feed the request directly.

## Things worth knowing

- **The spool file can no longer be renamed from the handler on Windows.** The stream stays open while the handler runs, and our `FileShare()` never includes `FILE_SHARE_DELETE`, so `MoveFileW` fails with a sharing violation. It worked before because the stream was closed at that point. Ownership transfer now goes through `DeleteFileOnDestroy := false` (or by excluding the flag) and processing the file after the request. Documented at the event.
- **A stream which finalizes in its destructor completes late.** `TSynZipCompressor` writes its `Flush` and `.gz` trailer in `Destroy`, so its destination is only complete after the request, not during it. That is the direct cost of keeping the stream alive, and the two cannot both be true. If you want an opt-out, the flags byte is now full.
- Spool files need a sharing mode to be read back by their `InContent` name while the stream is open: the tests use `fmCreate or fmShareRead`.

## TPipeStream

You asked to validate that `TPipeStream` works here. It does now — but getting there turned up #543: `TSynEvent.WaitFor` could report a timeout on an event which had been set, which made `TPipeStream.Read` return 0 as if the data had ended. About one request in eight then failed as a 507 after a 30 s stall. With your fix in, the case is covered by a test: an 8 MB body through a 64 KB pipe into a consumer thread, on both server families, 10 consecutive runs green.

Worth keeping in mind regardless: a blocking pipe write happens on the read poll thread in `THttpAsyncServer`, so it stalls that thread's other connections for the upload duration. Noted at the event.
## Tests

`TNetworkProtocols` `HttpBodyDownload` grew three cases, run against both server families: a self-deleting spool, a spool kept by `DeleteFileOnDestroy := false` (checked after server shutdown, and verified to go red without the fix), and a non-file stream read back from `InContentStream`. The existing spool path additionally asserts that `InContentStream` is present and rewinded, and a fourth case pipes the body to a consumer thread. The local `WaitDeleted` helper is gone, using your new `TSynTestCase.WaitDeleted`.

## Gate

Rebased on current master. FPC 3.2.2 aarch64 Linux: `TNetworkProtocols` 0/174246, `TTestCoreBase` 0/118823681, `TTestCoreProcess` 0/14841287. Delphi 13 Win32 and Win64: 0/448 each. The `HttpBodyDownload` block alone was run repeatedly to check for flakiness.

The two review rounds behind this: a self-deleting stream whose `DeleteFileOnDestroy` was false at setup time was still deleted by the server, contradicting the class contract; the `InContentStream` rewind raised on a non-seekable stream not descending from `TStreamWithNoSeek`, aborting the connection with no response; and `THttpServerRequest.Recycle` kept a pointer to the released stream.

### DIFF
--- a/src/core/mormot.core.os.pas
+++ b/src/core/mormot.core.os.pas
@@ -3451,6 +3451,31 @@ type
     function Write(const Buffer; Count: Longint): Longint; override;
   end;
 
+  /// file stream which deletes its own file once the instance is released
+  // - the temporary file lifetime is therefore bound to this instance, with no
+  // additional tracking needed by the caller
+  // - used e.g. as TOnHttpServerBodyDownload spool file: the request body is
+  // written into it, is available during the request process, then vanishes
+  // with the stream - this is the equivalent of nginx
+  // "client_body_in_file_only clean" (whereas "on" would keep the file)
+  // - set DeleteFileOnDestroy := false e.g. once the file has been renamed or
+  // moved by the process, so should not be deleted any more
+  TFileStreamEventuallyDelete = class(TFileStreamEx)
+  protected
+    fDeleteFileOnDestroy: boolean;
+  public
+    /// open or create the file, which will be deleted by Destroy
+    // - Mode is typically fmCreate, but a spooled file meant to be read back
+    // by its name while this instance is still open needs a sharing mode, e.g.
+    // fmCreate or fmShareRead (which is mandatory on Windows)
+    constructor Create(const aFileName: TFileName; Mode: cardinal); reintroduce;
+    /// close the handle, then delete the file unless DeleteFileOnDestroy=false
+    destructor Destroy; override;
+    /// set to false to keep the file after this instance is released
+    property DeleteFileOnDestroy: boolean
+      read fDeleteFileOnDestroy write fDeleteFileOnDestroy;
+  end;
+
 /// a wrapper around FileRead() to ensure a whole memory buffer is retrieved
 // - expects Size to be up to 2GB (seems like a big enough memory buffer)
 // - on Windows, will read by 16MB chunks max to avoid ERROR_NO_SYSTEM_RESOURCES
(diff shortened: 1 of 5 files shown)


landrix:
Rebased on current master and reworked:

- adopted your `TSynTestCase.WaitDeleted` and the `DoHttpBodyDownload` refactoring (`srv.SockPort`, `hosthead`) - the two conflicts were only the var block and the final `WaitDeleted` call
- added the `TPipeStream` coverage which I had left out before, now that #543 is fixed: 8 MB through a 64 KB pipe into a consumer thread, 10 consecutive runs green on both families. Its timeouts are bound on purpose, since they default to INFINITE and a regression should fail rather than hang the suite.

Gate on the rebased branch: FPC aarch64 `TNetworkProtocols` 0/174246, `TTestCoreBase` 0/118823681, `TTestCoreProcess` 0/14841287; Delphi Win32 and Win64 0/448 each.

Thanks for the #543 fix - and for extending it to the plain POSIX RtlEvent path, which I had not looked at.

synopse (approved):
Thanks again.

PR #545, opened 2026-08-19
ex: new http-server-upload sample streaming multipart uploads to disk

### LINKED ISSUE #292: Is there any future plan to support HTTP Server side File Upload?

landrix:
Closes #292

`OnBodyDownload` (#541, #544) and `THttpMultiPartDecoder` (#505) are both in, but
no sample shows them together - so #292 currently leads to two classes and no
entry point.

`POST /upload` returns a `TFileStreamEventuallyDelete` from `OnBodyDownload`: the
body is spooled to a temp file that vanishes with the request, and the decoder
copies each section straight to its destination. Any other URI returns `nil` and
keeps the in-memory path. `GET /` serves a form to try it from a browser.

Peak RSS (`VmHWM`) for one 400 MB upload, Linux/FPC: **89 MB** through the spool
stream, **406 MB** with `OnBodyDownload` returning `nil`.

Sample code gets copied, so the tricky parts are handled and commented:

- `TemporaryFileName()` for spool and staging names - `OnBodyDownload` runs on the
  event loop threads
- staging in a `.staging` sub-folder of the destination: same file system, but out
  of the served folder if the process is killed (swept at startup). A shared
  `<name>.part` breaks when two sections carry the same file name
- publish = rename first (atomic replace on POSIX), delete + retry only on failure
  (`RenameFile()` is `MoveFileW()` without `MOVEFILE_REPLACE_EXISTING`). Deleting
  upfront would lose the old file when the new one can't be put in place
- file name: path stripped, then `''`, `.`, `..`, trailing `.`/space and DOS device
  names refused - device check on the part before the *first* dot, otherwise
  `con.txt.txt` slips through
- media type checked separately: `CreateFromContentType()` only looks for a
  `boundary`, so `text/plain; boundary=x` would decode
- section count capped; `--maxsize` bounded (`MaximumAllowedContentLength <= 0` =
  no limit); errors echo no server paths or exception text
- `EHttpMultiPart` -> 400, not 500

README lists what the sample does *not* do: no auth, listens on all interfaces,
spool defaults to the world-readable temp folder, no atomic publish of a whole
set, and the decode still holds one server thread.

Verified: warning-free with FPC 3.2.2 (aarch64-linux) and dcc64 37.0. Runtime on
Linux: byte-identical round-trip to 400 MB, name matrix incl. the bypass attempts,
`text/plain; boundary=x`, 70 sections, failing rename (500, no leftover, target
intact), staging swept after SIGKILL, 413 / 400 / 404 paths.

Windows build is compiled only, not run - this machine is ARM64, so a failure
under Prism emulation could just as well be the emulator.

### DIFF
--- a/ex/README.md
+++ b/ex/README.md
@@ -53,6 +53,14 @@ It is a good example of what could be done to achieve the best performance with 
 
 ## Integrated and Advanced Samples
 
+### http-server-upload
+
+[This sample](./http-server-upload) is a stand-alone HTTP server receiving `multipart/form-data` file uploads without ever buffering the request body in memory.
+
+It combines `THttpServerGeneric.OnBodyDownload`, which spools the incoming body into a temporary file, with `THttpMultiPartDecoder`, which walks that spool file section by section - see [issue #292](https://github.com/synopse/mORMot2/issues/292).
+
+A 400 MB upload peaks at 89 MB of resident memory, against 406 MB via the default in-memory path.
+
 ### mvc-blog
 
 [MVC sample web application](./mvc-blog), publishing a simple BLOG.
--- a/ex/http-server-upload/README.md
+++ b/ex/http-server-upload/README.md
@@ -1,0 +1,117 @@
+# Streamed File Upload Server
+
+A stand-alone HTTP server which receives `multipart/form-data` uploads without
+ever holding the request body in memory - the answer to
+[issue #292](https://github.com/synopse/mORMot2/issues/292).
+
+## Running it
+
+Build `httpServerUpload.dpr` (or open `httpServerUpload.lpi` in Lazarus), then:
+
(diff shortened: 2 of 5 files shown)


synopse:
Thanks!

PR #547, opened 2026-08-22
Use an ABI-specific block register in _FreeMem

Moonbot-Tech:
Using `RSI` for `TSmallBlockType` is beneficial on System V because `RSI` is
caller-saved, so `_FreeMem` no longer needs to preserve it.

On Win64, however, `RSI` is nonvolatile just like `RBX`, so it still needs to
be saved and restored. In a compiled same-class allocation/free ring, the
caller also keeps its loop counter in `ESI`. Restoring `RSI` immediately before
return creates a dependency for the next iteration and caused a reproducible
Win64 regression.

This change selects the register at compile time, with no runtime branch:

- Win64 keeps `TSmallBlockType` in `RBX`;
- System V keeps it in `RSI`;
- the rare Win64 pending-bin path adapts `RBX` to the existing RSI helper
  through `R9`, without duplicating the helper.

Exact A/B on current upstream `master`, FPC 3.3.1, O3,
`FPC_X64MM + FPCMM_SERVER`, 12 paired interleaved medium runs of a
same-class 96-byte ring:

- upstream median: 19.131941 TSC ticks/operation;
- candidate median: 17.817627 TSC ticks/operation;
- candidate/upstream: 0.931303, or 6.87% faster.

Validation:

- Win64, stock FPC 3.2.2, full `TTestCoreBase`: 119,302,178 assertions,
  no failures;
- Win64 multithreaded `TTestCoreBase`: 114,936,181 assertions, no failures;
- all memory-manager performance cases completed with the expected digests;
- Linux probe binaries built from upstream and this branch are byte-identical
  (`SHA-256 4d8b827d2bc86c4beeab38e825e79d8a595d3fa3aa9fe5c79400d219550b71d2`).

### DIFF
--- a/src/core/mormot.core.fpcx64mm.pas
+++ b/src/core/mormot.core.fpcx64mm.pas
@@ -2491,7 +2491,7 @@ const
 
 function _FreeMem(P: pointer): PtrUInt;
   {$ifdef NOSFRAME} nostackframe; {$endif} assembler;
-asm     // P = rcx on Windows, P = rdi on SystemV; use rsi = TSmallBlockType
+asm     // P = rcx on Windows, P = rdi on SystemV; keep P in rcx on both ABIs
         {$ifndef MSWINDOWS}
         mov     rcx, P
         {$endif MSWINDOWS}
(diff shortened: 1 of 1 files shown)


synopse:
This is clearly too much verbose and too much $IFDEF everywhere.
Very error prone and non maintainable in the future.

Could you try with a FPC macro for the register name? I guess it could work.

synopse (changes requested):
Looks like if FPC macro are not expanded for the registers in the asm blocks. :(
We need to find something else, because I can't accept so many $ifdef in the middle of this unit, which is already very complex and stuffed with conditionals.

Please try just with a stack local variable to save rsi instead of pop/push. Since we have a stack frame anyway on Windows, it may be lighter. But note that optimizing for a specific case like the caller use of RSI is IMHO not a good idea. It would be very CPU and compiler specific.

To be fair, Win64 was never the premium target for performance, because it is so much slower and unstable than Linux for any demanding work (e.g. server process).

synopse:
I doubt changing rsi to rdi would change anything in practice.
Looks like a doubtful microbenchmarking.

Especially with your r9 trick to call GetSmallLastFreeBlock - xchg would be more suited perhaps.

Moonbot-Tech:
You were right. A symmetric A/B with otherwise identical callers showed that moving the saved register from RSI to RDI only moved the dependency: the RSI-indexed caller improved, while the RDI-indexed caller regressed.

I also tested a stack-slot version and an all-volatile-register version. Neither was a strict improvement across the neighboring tiny/small workloads. The xchg suggestion simplifies the rare helper adapter, but does not fix the underlying benchmark-specific trade-off.

I am closing this PR rather than presenting a caller-specific microbenchmark result as a general allocator improvement. Thank you for catching it.

PR #548, opened 2026-08-24
Postgresql+mysql additions

dkounal:
REPLACE INTO in mysql will delete the row and re-insert the row. But existing blobs' content in record, if any, will be lost.
So BoUpsert is created to use the following syntax: on duplicate key update nam=values(nam),val=values(val) 
at the same time postgresql can have now: insert into t (id,nam,val) values (?,?,?) on conflict (id) do update set nam=excluded.nam,val=excluded.val
In postgresql the key used is the primary key ID but this can be extend with changes and will not be compatible

Also in Postgresql the boInsertOrIgnore can work now as:
boInsertOrIgnore -> insert into … on conflict do nothing
boInsertOrReplace -> does not create now a bad formed sql

Please use only [19186a9] and [689f575]


synopse (approved):
sounds fine!
Thanks!

synopse:
Just some minor optimization needed.

Issue #411, opened 2025-11-10
WinARM Preliminary Support


synopse:
Windows on 64-bit ARM is a fact, with existing nice hardware (mostly laptops) around.

Goal of this issue is to see how far we could use FPC trunk to compile natively to the aarch64-win64 plaform.
We have very good support of arch64-linux support since years, validated nightly using our Tranquil IT LUTI Ampere server. ;)

We have identified and already fixed (circumvented) some compiler issues.
This entry could be used to mark new work in this direction.

landrix:
I am working on this issue on the side. We currently have two FPC blockers on
aarch64-win64, both reported with patches:

- WideString constants are addressed 0x4000 bytes too high (an addend treated as
  pages instead of bytes in the internal COFF linker) — this is what makes
  TTestCoreBase._TDynArray crash on the varOleStr variant key
  https://gitlab.com/freepascal.org/fpc/source/-/work_items/41859

- do_consttovar leaves a persistent register temp undeleted, so at -O2 the register
  holding a hoisted global address is reused mid-loop — this is the crash in
  TTestCoreBase.Utf8Slow
  https://gitlab.com/freepascal.org/fpc/source/-/work_items/41862

With both patched, TTestCoreBase runs to completion at plain -O2: 118.6M assertions,
26 failures, all in "Windows specific api".

synopse:
[First issue was a duplicated of my own report](https://gitlab.com/freepascal.org/fpc/source/-/work_items/41852) few minutes before yours, and with no solution. ;)

Hope we will stabilize the WinArm FPC target.
It is already in very good shape. 💃

landrix:
I specifically searched for existing issues. 🙈

landrix:
A third FPC blocker, patches attached as well:
https://gitlab.com/freepascal.org/fpc/source/-/work_items/41865

It is a recent Currency regression on trunk (four defects from one commit). With those
patches the aarch64-win64 test suite drops from 835 to 167 failures, and TDecimal128
passes: TDecimal128.ToCurr is an inline function returning Currency, which was the
single largest source of failures.

landrix:
Status update on aarch64-win64, with a native FPC 3.3.1 trunk compiler on a
Snapdragon X Elite. The regression suite now runs to completion:

    Total assertions failed for all test suits: 4 / 195,687,177

All four are `SELECT datetime(current_timestamp,'localtime')` returning NULL,
[...]

PR #550, opened 2026-08-24
tests: tolerate a SQLITE_OMIT_LOCALTIME sqlite3 library

landrix:
Follow-up to your suggestion in #411.

`datetime(...,'localtime')` returns NULL when the sqlite3 library was built
with `SQLITE_OMIT_LOCALTIME`, which is what `FOR_WIN10=1` in SQLite's
`Makefile.msc` does - the official win-arm64 binaries are built that way, so
`TTestSQLite3Engine` failed in all four Sqlite suites on Windows on ARM.

Rather than skipping the check, this asks the library: the void result is
accepted only when `sqlite_compileoption_used('OMIT_LOCALTIME')` confirms the
feature was left out, so a real failure is still reported.

With this, aarch64-win64 is green: `0 / 195,372,996` assertions in 1m15,
using FPC 3.3.1 trunk plus the four compiler fixes tracked in #411.
Unchanged on x86_64-win64, where the check takes the first branch as before.

### DIFF
--- a/test/test.orm.sqlite3.pas
+++ b/test/test.orm.sqlite3.pas
@@ -537,7 +537,12 @@ begin
   s := Demo.ExecuteNoExceptionUtf8('SELECT datetime(current_timestamp);');
   check(s <> '', 'datetime');
   s := Demo.ExecuteNoExceptionUtf8('SELECT datetime(current_timestamp,''localtime'');');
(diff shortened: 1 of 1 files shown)


synopse (approved):
Fair enough.

PR #551, opened 2026-08-24
core: support the FPC Unicode RTL on POSIX targets

Moonbot-Tech:
Follows the discussion in #527.

An FPC compiler with UnicodeString as the default string type (the `-dUNICODERTL` RTL mode) used to hit the explicit guard in `mormot.core.os.posix.inc`. This PR keeps the application-facing `string`/`TFileName` Unicode and transcodes to UTF-8 exactly at the raw POSIX boundaries (files, directories, `dlopen`, process arguments, environment, recursive search) — the same pattern the FPC Unicode RTL itself applies around its own syscalls.

**Safety:** every new branch is dead unless FPC defines `UNICODE`, so Delphi and default 8-bit FPC compile byte-for-byte identical code. To keep the conditionals maintainable they are synthesized once in `mormot.defines.inc` (`FPCUNICODE` / `FPCUNICODEPOSIX`), and the code carries plain `{$ifdef}`s only.

**Verification:** vanilla FPC main (`-dUNICODERTL`, Linux x86-64) compiles the eleven core units up to `mormot.core.json`; our Delphi-compatible FPC fork compiles Win64+Linux and runs the full mORMot test suite as its qualification battery (several hundred million assertions per run).

Two small gaps remain in *vanilla* FPC itself (not in mORMot, and not blocking this PR): the explicit `RawByteString(Variant)` cast of the Delphi-UNICODE arm of `VariantToUtf8` is not accepted yet, and a constant concatenation in an `external name` clause ICEs under unicodestrings (`LIBC_SUFFIX`). Both need one-line local workarounds on vanilla today; we plan to report both upstream to FPC with fixes, and the mORMot code stays untouched since it is correct for Delphi.

Scope is POSIX, matching the question in #527; the FPC+UNICODE combination on Windows needs a further pass (we have it working in our product line) and can follow as a separate PR.

### DIFF
--- a/src/core/mormot.core.json.pas
+++ b/src/core/mormot.core.json.pas
@@ -8007,7 +8007,9 @@ begin
       Ctxt.Info.Cache.Engine.Utf8BufferToAnsi(Ctxt.Value, Ctxt.ValueLen, Data^);
 end;
 
-procedure _JL_String(Data: PString; var Ctxt: TJsonParserContext);
+procedure _JL_String(
+  Data: {$ifdef UNICODE}PUnicodeString{$else}PAnsiString{$endif};
+  var Ctxt: TJsonParserContext);
 begin
   if Ctxt.ParseNext then
     Utf8DecodeToString(Ctxt.Value, Ctxt.ValueLen, Data^);
--- a/src/core/mormot.core.os.posix.inc
+++ b/src/core/mormot.core.os.posix.inc
@@ -8,6 +8,7 @@
 {$ifdef FPC}
 
 uses
+  cwstring,
   baseunix,
   unix,
   unixcp,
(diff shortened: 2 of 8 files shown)


src/core/mormot.core.json.pas
```diff
@@ -8007,7 +8007,9 @@ procedure _JL_RawUtf8(Data: PRawByteString; var Ctxt: TJsonParserContext);
       Ctxt.Info.Cache.Engine.Utf8BufferToAnsi(Ctxt.Value, Ctxt.ValueLen, Data^);
 end;
 
-procedure _JL_String(Data: PString; var Ctxt: TJsonParserContext);
+procedure _JL_String(
+  Data: {$ifdef UNICODE}PUnicodeString{$else}PAnsiString{$endif};
```
synopse: Isn't it superfluous?

src/core/mormot.core.os.posix.inc
```diff
@@ -8,6 +8,7 @@
 {$ifdef FPC}
 
 uses
+  cwstring,
```
synopse: Should be at project level.

src/core/mormot.core.os.posix.inc
```diff
@@ -35,15 +36,124 @@ uses
...
+// Moon Compiler builds the FPC RTL with Unicode String/TFileName on POSIX.
+// Keep mORMot's application-facing Unicode type and transcode at raw APIs.
 
 function FpOpenA(path: PAnsiChar; mode: cint): cint; inline;
 begin
   result := FpOpen(path, mode);
 end;
 
+function PosixDlopen(const FileName: TFileName; Flags: cint): pointer; inline;
+var
```
synopse: all those create an hidden try..finally because of the local raw RawUtf8 variable(s).

src/core/mormot.core.os.posix.inc
synopse: This Pull Request breaks the whole Delphi for POSIX compatibility.

synopse:
Good starting point but I need to rewrite almost every part of it.
I guess I will do it directly in the source.

The fact that you will eventually validate it is a great help.

src/core/mormot.core.rtti.fpc.inc
```diff
@@ -470,7 +473,23 @@ begin
     ent := tab^.Field[i];
     with result[i] do
     begin
-      TypeInfo := DeRef(ent^.FieldType);
+      info := DeRef(ent^.FieldType);
```
synopse: I guess it is needed for Delphi compatibility?

synopse:
I guess I included all your fixes and feedback.
Please check, verify and report.
:)

Which version/commit of FPC trunk do you use for testing?
The -dUNICODERTL seems not stable on the trunk I tested.

Moonbot-Tech:
We validated your master (a3c49ad) against vanilla FPC trunk `551ab370bb4b8241f77e91207a7f7db1997a1cdb` (2026-08-23 — the same trunk we used when testing this PR), Linux x86_64, `-dUNICODERTL`: 16 core/net units up to and including mormot.core.interfaces compile cleanly (base, os, unicode, text, datetime, rtti, buffers, data, variants, json, log, search, mustache, collections, threads, interfaces). The integration looks right — none of our fixes got lost in the rewrite.

About "-dUNICODERTL seems not stable on the trunk": you are most likely hitting the vanilla `Fatal: Internal error 2015092701`. It is neither your code nor our patch: in trunk's `compiler/widestr.pas` the `unicode2ascii()` branch for a UTF-8 codepage has been unimplemented since 2015 (`internalerrorproc(2015092701)` behind a "can't implement that here" comment), and any constant concatenation inside an `external`/`public name` clause wakes it up under `{$modeswitch unicodestrings}` + `{$codepage utf8}`. In your tree exactly two places trigger it — `mormot.lib.static.pas` (`_PREFIX + 'pas_malloc'`) and `mormot.net.sock.posix.inc` (`'gethostbyname' + LIBC_SUFFIX`) — which is why mormot.core.perf and mormot.net.sock fail while everything else compiles.

Minimal 10-line reproducer, no mORMot involved:

```pascal
{$mode delphi}
{$modeswitch unicodestrings}
{$codepage utf8}
unit icec;
interface
const XSUF = '2';
function gettimeofday: integer; cdecl;
  external 'c' name 'gettimeofday' + XSUF;
implementation
end.
```

`ppcx64 -O1 icec.pp` → `icec.pp(8,42) Fatal: Internal error 2015092701`.

We plan to report it to the FPC gitlab. A workaround on your side would be to expand the two concatenations into pre-built constants under ifdef — but properly speaking this is a compiler bug, not yours.

synopse:
And not the other constants like `'send' + LIBC_SUFFIX` in mormot.net.sock.posix.inc? weird.
I am sure it is worth reporting it to FPC gitlab. At least plain ASCII-7 chars are very easy to convert on the fly - if UTF-8 is out os scope.

Meanwhile:
https://github.com/synopse/mORMot2/commit/a4cc1eccd

synopse:
In fact the pre-computed constant trick is not enough.

I have another issue:
`function TVirtualStringList.Get(Index: integer): string; override;`
can't be compiled in mormot.core.unicode.pas because its parent TStrings.Get() uses string = AnsiString on my trunk.
I guess it is because my Lazarus may have not be compiled with `-dUNICODERTL`.

Moonbot-Tech:
On "why not `'send' + LIBC_SUFFIX`" — they all trip it; a `Fatal` simply stops the compilation at the first live site. The first live one is `gethostbyname` at line 355: the 244/281 sites sit under the disabled `SOCKDIRECTSYSCALL` and never reach the compiler. Remove 355 and `socket` at 358 fails next, and so on through all ~20 concatenations in the file.

Why the pre-computed constant trick cannot help: the ICE is not in the concatenation itself but in converting the resulting wide constant into the byte name of the symbol. Under `{$codepage utf8}` + `unicodestrings`, concatenating two ansi literals produces a *wide* constant (a single literal stays ansi — that is why declarations without the suffix compile fine), and when the compiler converts it into the linker name it enters `unicode2ascii()` with the UTF-8 codepage — the branch unimplemented since 2015. `external clib name LIBC_GETHOSTBYNAME` with a pre-built constant raises exactly the same ICE (verified with an isolated reproducer) — that is why a4cc1ec could not help and reverting it was right. The only local workaround would be whole per-combination literals under ifdef (`'gethostbyname@GLIBC_2.2.5'` as one piece), which is ugly times twenty sites. The right place for the fix is FPC itself — an ASCII-7 fast path, exactly as you said. We will report it to the FPC gitlab with the 10-line reproducer.

On `TVirtualStringList`: your guess is right, the mORMot code is fine. We compile `mormot.core.unicode` from your current master (faf738e) cleanly — on a stand where the RTL itself was rebuilt with `make rtl OPT='-O1 -dUNICODERTL'` using the same trunk compiler: then `TStrings.Get` returns UnicodeString and the override matches. The RTL and the application must be built in the same string mode — a default RTL build (the one Lazarus ships with) gives `string = AnsiString`, hence the error you saw.

For completeness: on faf738e all 16 core/net units are green for us with the same matrix as before.

synopse:
So I guess we can close this issue.

Thanks for sharing!

PR #553, opened 2026-08-26
mm: keep the memory manager installed at process exit

Moonbot-Tech:
The finalization uninstalled the MM and released its memory while units that finalize later (objpas among them) still hold managed values from this heap: their decrefs then hit released pages through the restored manager.

Measured: `SetResourceStrings` + normal exit = AV in `ObjPas.FinalizeResourceTables` (objpas finalizes after this unit — verified with breakpoints).

The OS reclaims everything at process exit anyway, so the manager now stays installed; the historical teardown is kept under `FPCMM_UNINSTALL_AT_EXIT` for allocator debugging.

### DIFF
--- a/src/core/mormot.core.fpcx64mm.pas
+++ b/src/core/mormot.core.fpcx64mm.pas
@@ -151,6 +151,16 @@ unit mormot.core.fpcx64mm;
 // try "cmp" before "lock cmpxchg" for old processors with huge lock penalty
 {.$define FPCMM_CMPBEFORELOCK}
 
+// restore the previous MM and release all its memory at unit finalization
+// - this was the historical behavior, but it is unsafe: units initialized
+// before this one finalize AFTER it (e.g. objpas, pulled in by this very
+// unit's mode), so their late managed-value finalizers would access an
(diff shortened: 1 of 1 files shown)


synopse (changes requested):
This would disable the whole FPCMM_REPORTMEMORYLEAKS feature.

I usually run everything with FPCMM_REPORTMEMORYLEAKS and have no leak reported.
So your "objpas may allocate before" statement is dubious.

synopse:
I guess my commits above (and below) may be a better fix for this issue: it isolates the root cause.

synopse:
But I have added a specific conditional FPCMM_FULLCLEANUP to disable the new default behavior.

Moonbot-Tech:
Thank you — your solution is better than our original PR. We initially overlooked that simply keeping the MM installed until process exit would effectively disable FPCMM_REPORTMEMORYLEAKS.
The dedicated FPCMM_FULLCLEANUP switch preserves both full cleanup and leak reporting, so no further changes are needed from our side. Thanks for the fix.

Issue #509, opened 2026-08-01
Let a server handler supply its own ContentStream (stream a response body without buffering it)


landrix:
`THttpServerRequestAbstract.OutContent` is a `RawByteString`, so a handler has to build the
whole response body in memory before returning. For large payloads that's RAM = body size ×
concurrent requests.

The only non-buffering way out today is `SetOutFile` / `SetOutProgressiveFile`, i.e.
`STATICFILE_CONTENT_TYPE` with a **file name**. Fine when the payload happens to be a file on
disk. Not fine when it isn't.

My case: file contents are stored as 4 MiB chunks spread over sharded SQLite databases. To serve
a download I currently concatenate every chunk into one `RawByteString`. There is no file to
point `SetOutFile` at, and materialising a temp file per download just to hand it back to the
server is a lot of pointless I/O.

What struck me while digging is that the machinery is already there — it's just not reachable
from a handler. `THttpRequestContext` (mormot.net.http.pas) already has:

- `ContentStream: TStream`
- `ProcessBody()`, documented as "read up to MaxSize bytes **from ContentStream** or fContentPos
  into Dest" — so the send loop is already stream-based
- `RangeOffset` / `RangeLength` / `ValidateRange`, and `ContentFromFile` already returns
  `HTTP_RANGENOTSATISFIABLE`
- `rfContentStreamNeedFree` for ownership
- `rfProgressiveStatic`, which shows the server can already wait for a stream to grow

`ContentFromFile` is essentially "open a `TFileStream`, assign it to `ContentStream`, validate
the range". Everything past that point is stream-based and file-agnostic. So this isn't a request
for new plumbing, it's a request for an entry point: let a handler assign its own `TStream` the
same way `ContentFromFile` assigns one, and let the existing range/send path do the rest.

Something along these lines:

```pascal
// THttpServerRequestAbstract
procedure SetOutStream(aStream: TStream; aOwned: boolean;
  const aContentType: RawUtf8 = ''; aContentLength: Int64 = -1);
```

Same semantics `SetOutFile` has today: the server picks up the stream, honours `Range`, emits
`Content-Range` + 206, and 416 on an unsatisfiable one. Ownership through
`rfContentStreamNeedFree`. If the stream isn't seekable, I'd expect range support to degrade to
`Accept-Ranges: none` rather than fail.

About `aContentLength = -1`: I'd define it as "derive the length from
`aStream.Size - aStream.Position`" — which is exactly what `ContentFromFile` does with the file
[...]

PR #554, opened 2026-08-26
Proposal: let a handler supply its own response body stream - see #509

landrix:
This is a **proposal**, not a finished feature waiting to be merged. #509 has had no
answer yet, and the API shape is your call - I wrote this so there is something
concrete to react to, and I am happy to throw any part of it away.

The send loop is already stream based (`ContentStream`, `ProcessBody`,
`ValidateRange`, `rfContentStreamNeedFree`), so this adds an entry point rather
than new plumbing.

Note that this builds on the `Range:` overflow fix you merged as b61d9db4d: the
feature relies on its `ProcessBody()` abort, since a handler-supplied stream can
end short far more easily than a file can.

## API

```pascal
// THttpServerRequestAbstract
function SetOutStream(aStream: TStream; aOptions: THttpOutStreamOptions = [];
  const aContentType: RawUtf8 = ''; aContentLength: Int64 = -1): cardinal;
```

Returns `HTTP_SUCCESS`, like the other `SetOut*` methods, so a handler can write
`result := Ctxt.SetOutStream(...)`.

- `aContentLength = -1` computes `aStream.Size - aStream.Position`, like a file
  would use its size; that position is also the origin of any `Range:`
- `hosOwned` hands the stream to the server, which frees it once the response is
  sent or the request aborted
- `hosNoRange` declares up front that this stream serves no `Range:`, for a source
  which can only be read forward; it is also set automatically when the stream
  turns out not to be seekable
- a body of *unknown* length stays out of scope, as I wrote in the issue: the
  send loop always emits `Content-Length:` and has no chunked response encoding

## How it is wired

The actual work is a new `THttpRequestContext.ContentFromStream()`, sitting right
next to `ContentFromFile()` and doing the same things: set `ContentLength`,
validate the range, seek, assign `ContentStream`, set `rfContentStreamNeedFree`,
return `HTTP_SUCCESS` or `HTTP_RANGENOTSATISFIABLE`. `THttpServerRequest.
ProcessOutStream` is then a ~20 line wrapper called from `SetupResponse`, mirroring
`ProcessStaticFile`. Both socket server families share that path, so both stream it.

- ranges behave like a file: 206 with `Content-Range:`, 416 when unsatisfiable
- a stream which can not seek - a `TStreamWithNoSeek` such as `TPipeStream`, or
  one whose `Position`/`Size` raise - is served sequentially with
  `Accept-Ranges: none`, and a `Range:` request on it is ignored, so it answers
  200 with the whole body rather than a 206 without a `Content-Range:`
- servers which can not stream an in-process `TStream` - `THttpApiServer`, where
  http.sys sends from the kernel, and REST-over-WebSockets answers - call
  `OutContentStreamToBuffer`, which reads it into `OutContent`. A handler stays
  portable and only loses the memory benefit there, mirroring how `OnBodyDownload`
  falls back when it returns nil. Rejecting outright would have been the other
  option; I picked the one that does not break a handler moved between server
  classes. For http.sys this happens inside `SendResponse`, so that an
  `OnBeforeRequest` answering early cannot bypass it.

## Ownership and error paths

The stream is owned by the request until `SetupResponse` hands it to the context,
and by the context afterwards. It is released on every path I could find:
`ProcessDone` after sending, `Recycle` on a reused `THttpAsyncServer` connection,
the new `THttpServerRequestAbstract.Destroy` if the connection dies first, a
second `SetOutStream` call, and the 416 rejection.
`THttpAsyncServer.AsyncResponse` discards a pending stream before assigning its
delayed content, otherwise the stale stream would win in `SetupResponse`.

Two cases are worth calling out, because I got both wrong at first:

- a handler which raises *after* `SetOutStream` had its body sent as the 500
  response, since `CompressContentAndFinalizeHead` uses `ContentStream` and never
  looks at the error page in `OutContent`. An error response now always discards a
  pending stream.
- a handler returning a non-200 status with its own stream body had that body
  range-processed - a `Range:` request could truncate a 404 body, or replace it
  with the generated 416 page. Range handling is now gated on `StatusCodeIsSuccess`.

Note: `THttpServerRequest.Destroy` previously did not call `inherited` (it was
void); it does now. An external descendant overriding `Destroy` without calling
`inherited` would leak - none exists in this repository.

## Tests

`TNetworkProtocols.DoHttpOutStream` runs the same steps against **both**
`THttpServer` and `THttpAsyncServer`: a 3MB body from an owned stream, a range
request with its `Content-Range:` header, an unsatisfiable range, an explicit
shorter length, a replaced stream, a caller-owned stream which must survive, a
handler which raises after `SetOutStream`, a non-200 response with a range, a
stream supplied already positioned (with and without a range), a partially
consumed `TStreamWithNoSeek`, a stream whose `Position` raises without it being a
`TStreamWithNoSeek`, and one which does answer `Position`/`Size` but raises on the
actual `Seek()`.

Those last two are deliberately separate, because they exercise different guards:
the first is caught in `SetOutStream` and downgrades the stream to
`Accept-Ranges: none`, the second is caught in `ContentFromStream` and answers 416.
Keeping them apart also keeps the test compiler independent - `GetPosition` is
virtual under FPC only, so a helper stream which raises on `Seek(0, soCurrent)`
alone would take one path under FPC and the other under Delphi.

Ownership is counted, not assumed: the test streams increment a counter in their
destructor, so a leak or a double free fails an assertion. Every claim was
verified by reverting the corresponding code and confirming the matching
assertions fail - on both server families separately.

## What was measured, and what was not

I develop on ARM64, where x64 only runs under emulation, so everything needing a
native x64 result was run on separate hardware (Windows 11, i9-14900K, FPC 3.2.2
x86_64-win64, Delphi 13 `dcc64` and `dcc32`).

- FPC 3.2.2, aarch64-linux and x86_64-win64: `TNetworkProtocols` green and stable
  over repeated runs.
- Delphi 13, Win64 **and** Win32, actually executed: `HTTP` and `_SocketIO` green
  over 30 runs each. Win32 matters here because the whole path is `Int64` sized.
- `THttpApiServer` (http.sys), exercised directly rather than through the suite:
  a 1MB stream arrives byte identical, and a 3MB one - above
  `HttpContentFromFileSizeInMemory` - answers 500 with an empty body instead of a
  truncated one.
- REST-over-WebSockets, the same two cases through `TWebSocketServerRest`, same
  result. That is the path `OutputToFrame` gained, and it had never been executed
  before.
- Still not covered by a test: HTTP pipelining, `Recycle` cleanup of a pending
  stream, and a connection dying between the handler and `SetupResponse`. I traced
  them by reading the code and believe they are sound, but no assertion proves it.

That x64 round also caught a real defect, fixed in the current commit: a test
helper stream raised on every `Seek()`, including the harmless `Seek(0, soCurrent)`
that Delphi's `TStream.Position` uses, so `/failseek` took a different branch per
compiler - 30 out of 30 Delphi runs red, green under FPC throughout. The shipped
code was correct on both; the test expectation was not.

## Open questions, where I would follow your preference

1. ~~**Non-seekable detection.**~~ Settled by your review: automatic detection
   stays, and `hosNoRange` is there for a handler which already knows.
2. ~~**The other backends.**~~ Settled: they buffer into `OutContent`.
3. **206 status selection.** `SetupResponse` picks 206 from `rfWantRange` alone,
   before knowing whether a partial response actually happens. That never shows
   with files, since a file always validates the range or fails with 416; it
   surfaced here for the non-seekable and non-200 cases, which I handled by
   clearing `rfWantRange`. If you would rather have the status follow `rfRange`,
   that is a one line change in `SetupResponse` and both workarounds can go.
4. **416 without `Content-Range: bytes */size`.** Unchanged from `ContentFromFile`,
   so this proposal does not alter it - mentioning it only because RFC 9110
   recommends the header.

My use case, for context: file contents stored as 4MB chunks across sharded
SQLite databases, so there is no file to point `SetOutFile` at, and a temporary
file per download would be pointless I/O.

### DIFF
--- a/src/net/mormot.net.http.pas
+++ b/src/net/mormot.net.http.pas
@@ -350,6 +350,20 @@ type
     rfAsynchronous,
     rfProgressiveStatic);
 
+  /// define THttpRequestContext.ContentFromStream() behavior, as supplied by
+  // THttpServerRequestAbstract.SetOutStream()
+  // - hosOwned will release the stream once the response has been sent, or the
+  // request aborted - otherwise the caller does own it, and should keep it
+  // valid until the response has been sent, i.e. after the handler returned
+  // - hosNoRange won't serve any 'Range: ' request from this stream, but
+  // return an 'Accept-Ranges: none' header instead - note that a
+  // TStreamWithNoSeek, or a stream raising on Position/Size, is detected as
+  // such by itself, so this option is only needed for other classes
+  THttpOutStreamOption = (
+    hosOwned,
+    hosNoRange);
+  THttpOutStreamOptions = set of THttpOutStreamOption;
+
   /// define THttpRequestContext.ProcessBody response
   // - hrpSend should try to send the Dest buffer content
   // - hrpWait is returned in rfProgressiveStatic mode to wait for more data
@@ -540,6 +554,17 @@ type
     // - if CompressGz is set, would also try for a cached local FileName+'.gz'
     // - returns HTTP_SUCCESS, HTTP_NOTFOUND or HTTP_RANGENOTSATISFIABLE
     function ContentFromFile(const FileName: TFileName; CompressGz: integer): integer;
+    /// initialize ContentStream/ContentLength from a given TStream
+    // - as supplied by THttpServerRequestAbstract.SetOutStream() to a handler
+    // - aStart is the stream position of the first content byte, and the
+    // origin of any 'Range: ' request; aLength is the content size in bytes
+    // - hosOwned does set rfContentStreamNeedFree, i.e. the stream is released
+    // by ProcessDone once the response has been sent
+    // - hosNoRange serves no range: an 'Accept-Ranges: none' header is
+    // returned instead, and any Range: request is ignored
+    // - returns HTTP_SUCCESS or HTTP_RANGENOTSATISFIABLE
+    function ContentFromStream(aStream: TStream; aStart, aLength: Int64;
+      aOptions: THttpOutStreamOptions): integer;
     /// uncompress Content according to CompressContentEncoding header
     procedure UncompressData;
     /// check RangeOffset/RangeLength against ContentLength
@@ -896,6 +921,10 @@ type
     fInContent: RawByteString;
     fInContentStream: TStream;
     fOutContent: RawByteString;
+    fOutContentStream: TStream; // from SetOutStream()
+    fOutContentStreamLength: Int64;
+    fOutContentStreamPos: Int64; // stream position when SetOutStream() was set
+    fOutContentStreamOpt: THttpOutStreamOptions;             // 8-bit
     fConnectionID: THttpServerConnectionID;                  // 64-bit
     fConnectionFlags: THttpServerRequestFlags;               // 8-bit
     fAuthenticationStatus: THttpServerRequestAuthentication; // 8-bit
@@ -913,6 +942,9 @@ type
     function GetRouteValue(const Name: RawUtf8): RawUtf8;
       {$ifdef HASINLINE} inline; {$endif}
   public
+    /// finalize this instance
+    // - will release any pending SetOutStream() owned stream
+    destructor Destroy; override;
     /// prepare an incoming request from a parsed THttpRequestContext
     // - will set input parameters URL/Method/InHeaders/InContent/InContentType
     // - won't reset other parameters: should come after a plain Create or
@@ -1037,9 +1069,67 @@ type
     // status HTTP_NOTMODIFIED (304) if it did not change
     function SetOutContent(const Content: RawByteString; Handle304NotModified: boolean;
       const ContentType: RawUtf8 = ''; CacheControlMaxAgeSec: integer = 0): cardinal;
+    /// set the response body to be read from a TStream, with no buffering
+    // - the socket-based servers do send this stream by chunks from their own
+    // sending loop, with a constant memory usage, as they already do for a
+    // static file - so a handler can serve content which is not a local file,
+    // e.g. rebuilt from database blobs, without a whole RawByteString body
+    // - aContentLength = -1 computes aStream.Size - aStream.Position, as a file
+    // would use its size: the stream is sent from its current position, which
+    // is also the origin of any Range: - so don't move it after this call
+    // - supply an explicit aContentLength if Position/Size do raise on this
+    // stream: it is then sent sequentially, and is the only length we know
+    // - a body of unknown length is NOT supported, since the send loop always
+    // emits a Content-Length: and has no chunked response encoding
+    // - if aOwned is true, the stream is released by the server once the
+    // response has been sent, or the request aborted: a handler should not
+    // free it, nor keep any reference - if aOwned is false, it shall stay
+    // valid until the response has been sent, i.e. after the handler returned
+    // - 'Range: ' is served as for a file, i.e. 206 with a Content-Range: or
+    // 416 if unsatisfiable; a stream which can not seek - a TStreamWithNoSeek
+    // like TPipeStream, or one raising on Position/Size - is detected as such
+    // and gets an 'Accept-Ranges: none' header, ignoring any Range:
+    // - a stream which does answer Position/Size but raises on the actual
+    // seek can not be detected in advance, so it returns 416 for that range:
+    // supply hosNoRange for such a stream, to serve it sequentially instead
+    // - THttpApiServer (http.sys) and REST-over-WebSockets can not stream an
+    // in-process TStream: they read it into OutContent, so that a handler is
+    // portable, with no Range: support and no memory gain on those servers -
+    // but only up to HttpContentFromFileSizeInMemory (1MB on 32-bit, 2MB on
+    // 64-bit): a bigger body does answer 500 there, so raise that global
+    // variable if your handler may run on those servers
+    // - the body is not compressed, just like a static file is not
+    // - a stream supplying less bytes than announced aborts the response and
+    // closes the connection, once the headers have been sent
+    // - a stream raising in Read() is not catched by the sending loop: as for
+    // any ContentStream, it will abort the whole THttpAsyncServer write
+    // thread, so a handler stream should return 0 rather than raise
+    // - the length is mandatory: a stream with no Size and no aContentLength
+    // is rejected with HTTP_SERVERERROR, since the send loop always emits a
+    // Content-Length: and has no chunked response encoding
+    // - calling it twice does release any previously owned stream
+    function SetOutStream(aStream: TStream;
+      aOptions: THttpOutStreamOptions = []; const aContentType: RawUtf8 = '';
+      aContentLength: Int64 = -1): cardinal;
     /// append a new line of HTTP headers to the request output
     // - just a wrapper around AppendLine(fOutCustomHeaders, Args)
     procedure SetOutCustomHeader(const Args: array of const);
+    /// read any pending SetOutStream() body into the OutContent memory buffer
+    // - to be called by the servers which can not stream a response body, e.g.
+    // THttpApiServer (http.sys) or a REST-over-WebSockets answer
+    // - returns true on success, or if there was no pending stream at all
+    // - returns false if reading the stream failed, leaving a void OutContent:
+    // those servers track their status aside, so the caller sets its own error
+    function OutContentStreamToBuffer: boolean;
+    /// release any pending SetOutStream() body without reading it
+    // - if we do own it, e.g. once the response has been sent otherwise
+    procedure OutContentStreamDiscard;
+      {$ifdef HASINLINE} inline; {$endif}
+    /// the TStream supplied to SetOutStream() as response body, or nil
+    // - published as read-only, e.g. for the servers to detect that a body
+    // has been set even if OutContent is void
+    property OutContentStream: TStream
+      read fOutContentStream;
   published
     /// input parameter containing the caller URI
     property Url: RawUtf8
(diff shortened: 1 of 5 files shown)


src/net/mormot.net.http.pas
```diff
@@ -1037,9 +1058,48 @@   THttpServerRequestAbstract = class
...
+    /// read any pending SetOutStream() body into the OutContent memory buffer
+    // - to be called by the servers which can not stream a response body, e.g.
+    // THttpApiServer (http.sys) or a REST-over-WebSockets answer
+    // - returns true on success, or if there was no pending stream at all
+    // - returns false if reading the stream failed, leaving a void OutContent:
+    // those servers track their status aside, so the caller sets its own error
+    function OutContentStreamToBuffer: boolean;
+    /// release any pending SetOutStream() body without reading it
+    // - if we do own it, e.g. once the response has been sent otherwise
+    procedure OutContentStreamDiscard;
```
synopse, line 1126: make it inline

src/net/mormot.net.http.pas
```diff
@@ -4841,6 +4946,12 @@ function ToText(a: THttpServerRequestAuthentication): PShortString;
 
 { THttpServerRequestAbstract }
 
+destructor THttpServerRequestAbstract.Destroy;
+begin
+  OutContentStreamDiscard; // e.g. on a connection aborted before SetupResponse
```
synopse: perhaps just these lines are enough:
```
 if ifOutContentStreamOwned in fInternalFlags then
    FreeAndNilSafe(fOutContentStream);

**Comment by synopse** [comment]
1) About InheritsFrom(TStreamWithNoSeek) and booleans: why not just use a set?
It sounds simple, readable and expandable. [] as default then proper options.
For the two methods setting the TSTream.
But keep the TStreamWithNoSeek detection anyway because it is as expected.

2) Buffering the response could make sense - only if it is not too big in size.
Use the regular size limit for responses.

3) I am no RFC expert. But I guess 2xx means success and 4xx means error. Otherwise, the final code is less decisive.

4) I don't know. Perhaps in the future if someone complains about it.

File: src/net/mormot.net.http.pas [hunk 3]
```diff
@@ -5152,6 +5263,79 @@ function THttpServerRequestAbstract.SetOutContent(const Content: RawByteString;
...
+
+function THttpServerRequestAbstract.OutContentStreamToBuffer: boolean;
+var
+  tmp: TRawByteStringStream;
+begin
+  result := true; // nothing to do is a success
+  if fOutContentStream = nil then
+    exit;
+  // this server can not stream a response body: read it as a regular content
+  tmp := TRawByteStringStream.Create;
```
synopse: I guess a FastNewString + ReadAll() could make the trick since we have fOutContentStreamLength

src/net/mormot.net.server.pas
```diff
@@ -3376,6 +3378,31 @@ procedure THttpServerRequest.ProcessErrorMessage;
   fOutContentType := HTML_CONTENT_TYPE; // body = human friendly HTML message
 end;
 
+procedure THttpServerRequest.ProcessOutStream(var Context: THttpRequestContext);
+begin
+  // hand a SetOutStream() body to the context, as ProcessStaticFile() does
+  if fRespStatus <> HTTP_SUCCESS then
```
synopse: only accept 200 is perhaps too restrictive

test/test.net.proto.pas
```diff
@@ -4379,9 +4351,78 @@ procedure TNetworkProtocols.HTTP;
...
 type
+  // a stream which does count its own destruction, from the server threads
+  TCountedStringStream = class(TRawByteStringStream)
+  public
+    destructor Destroy; override;
+  end;
+
+  // a readable stream which raises on Size/Seek, and does NOT descend from
+  // TStreamWithNoSeek: SetOutStream() should detect it the hard way
+  TRaiseSeekStream = class(TStream)
```
synopse: you may have inherited from TRawByteStringStream and override only Seek/Write

test/test.net.proto.pas
```diff
@@ -4712,6 +4810,284 @@ procedure TNetworkProtocols.DoHttpFileRange(Sender: TObject);
...
+        srv.Free;
+      end;
+    end;
+  finally
+    FreeAndNil(outkept);
+    outdata := '';
+    TSynLog.Family.ExceptionIgnore.Remove(EHttpServer);
+  end;
+end;
+
```
synopse, line 5305: I like those tests. :)

synopse (changes requested):

synopse:
I am very pleased with the current state.

Should we merge?

synopse (approved):

landrix:
I'll let you know; I'm running a few more tests.

landrix:
Both hunks I flagged as unverified are now exercised, on native x64 hardware
(Windows 11, i9-14900K, FPC 3.2.2 x86_64-win64, Delphi 13 `dcc64` and `dcc32`).
I have updated the PR description accordingly.

**http.sys and WebSockets**, tested directly rather than through the suite, since
no test drives a `SetOutStream` body through either. Same two cases on both, with
`HttpContentFromFileSizeInMemory` at 2097152:

```
small status = 200
small length = 1048576 (expected 1048576)
small match  = TRUE
big status   = 500
big length   = 0
```

So the 1MB stream is buffered and arrives byte identical, and the 3MB one answers
500 with an empty body rather than a truncated one.

**Delphi**, executed for the first time rather than only compiled: `HTTP` and
`_SocketIO` green over 30 runs each, Win64 and Win32. Win32 was worth doing - the
whole path is `Int64` sized.

That run caught a real defect, which is why the branch moved. My `TRaiseSeekStream`
raised on every `Seek()`, including the `Seek(0, soCurrent)` that Delphi's
`TStream.Position` uses because `GetPosition` is only virtual under FPC. So
`/failseek` was caught by `SetOutStream` under Delphi and by `ContentFromStream`
under FPC - 30 out of 30 Delphi runs red, green under FPC throughout. The shipped
code was right on both compilers, and defensively so under Delphi; only my test
expectation was wrong.

Fixed by making the helper raise on `Position` under both compilers, so
`/failseek` now asserts the `SetOutStream` guard, and by adding `TNoJumpStream`
and `/failrange` for a stream which knows its position but cannot jump - that one
asserts the `ContentFromStream` guard and its 416. Both were checked by reverting
the guarded line and confirming exactly those assertions go red, on both server
families.

One aside, unrelated to this PR: `TWebSocketProtocol.SetEncryptKey` returns
silently when `aSettings = nil`, leaving encryption off entirely. A client built
by hand that way talks plaintext to an encrypting server and every frame is
dropped with `sprBadRequest`. It cost me a while to find; a hint in the doc
comment might save the next person the same detour.

landrix:
So, from my perspective, everything is ready for now.

landrix:
Heads-up: this moved again after your approval, so it deserves another look
rather than a silent merge.

Rebasing onto your TFTP work left three merge artefacts in my own commit -
`published  published` and two doubled comments in `test.net.proto.pas`. They
compiled, so nothing caught them. Removed, and I now scan the whole diff for
that pattern instead of only checking that the symbols survived.

Then I had the branch reviewed again, and it turned up four real defects that
two earlier rounds had missed. Each fix is red-tested: I reverted the guarded
line and confirmed exactly the matching assertions fail.

- `OutContentStreamToBuffer` let a `Read()` exception escape, against its own
  doc comment. On http.sys that is the bad one: `respsent` is already `true` at
  that point, so the outer handler sends no 500 and the client waits for a
  response that never comes. Now caught and reported as `false`.
- `ContentFromStream` overwrote a `ContentStream` that `ContentFromFile` had
  already set, without clearing `rfContentStreamNeedFree`. After
  `SetOutStream` + `SetOutFile` the file stream leaked and a caller-owned
  stream got freed by us. It now releases the previous owned stream and clears
  the flag.
- The `TUriRouter` and `OnBeforeRequest` paths test only `OutContent = ''` to
  decide a callback returned no body. A `SetOutStream` body is void by design,
  so a routed handler answering 404 from a stream got `'Wrong route'` and my
  error page instead of its own body - exactly what `ProcessOutStream` claims
  to prevent, and it did work through `OnRequest`, which is why my test missed
  it. Both checks now also look at the stream, through a new read-only
  `OutContentStream` property. There is a `/routed` test now.
- A non-2xx stream answer advertised `Accept-Ranges: bytes` while its `Range:`
  had just been deliberately ignored. It now says `none`.

Two more places where the code was fine but I had promised too much:
`SetOutStream` said a raising stream "aborts the response and closes the
connection" - it does not, a raising `Read()` takes down the whole
`THttpAsyncServer` write thread, as it would for any `ContentStream`. The doc
now says so and tells handlers to return 0 rather than raise. It also now
states the `HttpContentFromFileSizeInMemory` limit on http.sys/WebSockets,
which my own use case exceeds. And a stream with no measurable size and no
supplied length used to answer an empty 200 - it is now refused with 500, since
the send loop always emits a `Content-Length:`.

One report I did not act on: that ranges are processed for 201/3xx because
`StatusCodeIsSuccess` spans 200-399, while only 200 becomes a 206.
`ProcessStaticFile` does not check the status at all, so the stream path is
already stricter than the file path, and narrowing it back to `= 200` would
undo what you asked for in review. Happy to change it if you would rather have
that.

FPC aarch64-linux green and stable at 0/663 over repeated runs, Delphi 13
`dcc64` clean. The native x64 numbers in my previous comment predate these
changes; I can have that rerun if you want it before merging.

landrix:
The native x64 run I promised is done, against `e73c8c784` this time (Windows 11,
i9-14900K, FPC 3.2.2 x86_64-win64, Delphi 13 `dcc64` and `dcc32`).

**The case that actually mattered.** `OutContentStreamToBuffer` is what http.sys
and the WebSockets answer use, and it is what changed, so a stream raising in
`Read()` was tested end-to-end on both, with a timer around the call:

```
small status = 200
small length = 1048576 (expected 1048576)
small match  = TRUE
big status   = 500
big length   = 0
raise status = 500
raise length = 0
raise time   = 0 ms
```

Before the fix that last one hung until the client timed out, since `respsent`
was already `true` and the outer handler sent no 500. Now it answers
immediately. Same on `TWebSocketServerRest`.

**Delphi**, 20 runs per test case per bitness, 80 in total: no hit on any of
`failseek`, `failrange`, `routed`, `notfound none`, `buffer raise`,
`no length refused`, `previous owned stream released`, `not ours to free`.
`0 / 302` and `0 / 629` throughout, identical on Win32 and Win64.

The router path was also checked end-to-end with a separate program registering
the same handler twice, once as `OnRequest` and once via `Route.Get`: both
return 404 with the full stream body and `Accept-Ranges: none`.

**One observation I would rather report than sit on.** In 180 FPC runs,
`HttpFileRange` failed completely twice - the client got no answer at all on the
full 3MB GET, empty body - against 0 in 180 runs on `ca302d174`. Those
assertions are not part of this PR and the file path does not touch any of my
code. What my test does add is load: it transfers 3MB of its own and starts two
more servers, so the PR build runs four background tests where the base runs
three, and the ports do not collide (8893/8894 against 8895/8896). So this looks
like existing fragility surfacing under more parallel load rather than something
new - but 2 against 0 is not nothing, and at roughly 1% you would need a lot
more runs to tell those apart.

If you would rather not carry that extra load in the suite, I can drop
`DoHttpOutStream`'s payload from 3MB to something smaller - it only needs to
exceed the socket buffer to exercise the multi-chunk send path. I did not do it
on my own, since shrinking my test so another one flakes less seemed like your
call rather than mine.

synopse:
About the confused `TWebSocketProtocol.SetEncryptKey` 
please see
https://github.com/synopse/mORMot2/commit/57660dac2

PR #556, opened 2026-08-26
net: own and join TFTP connection threads before server teardown

Moonbot-Tech:
## Problem

`TTftpConnectionThread` was started from its constructor with `FreeOnTerminate = true`, then registered in a non-owning `fConnection` list only after the constructor returned.

This left two lifetime races:

1. A short-lived worker could finish and destroy itself before being added, leaving a dangling pointer in the server list.
2. During shutdown, the server only called `Terminate` and cleared the non-owning list. A worker could therefore keep using `fOwner` after the server and its state had been released.

This can result in a use-after-free during TFTP server teardown.

## Fix

- Construct connection threads suspended.
- Register each worker before `Start`.
- Make the server list own non-self-freeing workers.
- Reap finished workers before accepting a new request and from `OnIdle`.
- On teardown, terminate all workers and let the owning list destroy and join them before releasing server state.
- Preserve the existing timeout behavior of `TerminateAndWaitFinished`; the mandatory unbounded join happens only when the owner is actually destroyed.
- Release socket, context, and stream state only after `TThread.Destroy` has joined `Execute`.

## Regression test

A cross-platform `TNetworkProtocols.TFTPShutdown` test creates a worker that deliberately remains alive beyond a 1 ms shutdown timeout and then accesses scalar owner state. `Server.Free` must return only after the worker destructor has run.

Results:

- Exact pre-fix upstream: deterministic failure, `connection outlived its owner`.
- Delphi 12.2 Win64: PASS.
- FPC 3.2.2 Linux x86-64 O2: PASS.
- Real 65,537-byte TFTP/libcurl transfer with content comparison: PASS.

### DIFF
--- a/src/net/mormot.net.tftp.server.pas
+++ b/src/net/mormot.net.tftp.server.pas
@@ -93,6 +93,7 @@ type
     fLastSent: pointer;
     fLastSentLen: integer;
     procedure DoExecute; override;
+    function HasFinished: boolean;
   public
     /// initialize this connection
     constructor Create(const Source: TTftpContext; Owner: TTftpServerThread); reintroduce;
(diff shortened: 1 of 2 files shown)


synopse (approved):
I am not sure it would compile on older Delphi but I will check.

PR #558, opened 2026-08-27
Fix ValueVarToVariant managed value leak on NULL

Moonbot-Tech:
## Problem

`ValueVarToVariant(nil, ..., result)` directly changed `VType` to `varNull`.

If `result` already contained a string or another managed value, its reference was lost without clearing the Variant, leaking the value.

## Fix

Use `VarClearAndSetType(..., varNull)` for the NULL branch, as already done by the regular `ValueVarToVariant` path.

The existing `TTestCoreProcess.Variants` test now verifies that replacing a string Variant with NULL releases exactly one reference.

## Validation

- Pre-fix focused regression: the reference count was unchanged.
- FPC 3.2.2 Win32: focused regression passed.
- Delphi 12.2 Win64: focused regression passed.
- `TTestCoreProcess.Variants`: 197 assertions passed.

### DIFF
--- a/src/orm/mormot.orm.base.pas
+++ b/src/orm/mormot.orm.base.pas
@@ -3624,7 +3624,7 @@ var
 begin
   if Value = nil then
   begin
-    TSynVarData(result).VType := varNull;
+    VarClearAndSetType(variant(result), varNull);
     exit;
   end;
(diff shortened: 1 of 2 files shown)


synopse (approved):
Indeed!

Issue #401, opened 2025-10-28
Add TDebugFile.StackTrace methods


synopse:
Currently, the stack trace could be written to the log.

We would like to have a way to return the stack trace information from the current thread, in a cross-platform way, as convenient text.

synopse:
I guess we got what we needed.

Thanks! 🧑‍🍳

PR #559, opened 2026-08-31
core: add TDebugFile.StackTrace() methods - request #401

zen010101:
Implements #401: return the current thread stack trace as convenient text, with no exception involved.

## API

Two overloads on `TDebugFile`:

```pascal
class function StackTrace(skip: integer = 0; depth: integer = 0): RawUtf8; overload; static;
class procedure StackTrace(W: TTextWriter; skip: integer = 0; depth: integer = 0); overload; static;
```

`skip` = caller frames to ignore (0 = start at the caller); `depth` = maximum located frames (0 = 30, as the `TSynLogFamily.StackTraceLevel` default). The output format matches the existing TSynLog frame text, e.g.:

```
401545 test.lpr ZvbProbeLevel3 (54) 401919 test.lpr ZvbProbeLevel2 (93) 401939 test.lpr ZvbProbeLevel1 (98) 40197f test.lpr $main (104)
```

## Implementation notes

- One single capture implementation: `TSynLog.AddStackTrace` (FPC) now delegates to the shared code, so the log output and the API output cannot drift apart. Verified A/B on aarch64-linux: the `sllError` frame list is byte-identical (modulo the code addresses themselves, which moved with the sources).
- Capture uses the RTL `CaptureBacktrace()` on FPC, and the `RtlCaptureStackBackTrace()` API on Delphi + Windows; Delphi POSIX is not implemented yet (returns `''`), as in the previous code.
- The new functions are wrapped in `{$STACKFRAMES ON}` (as `TSynLog.Enter` already does): on x86_64 the RTL backtrace walks the rbp chain, so the *calling* code needs stack frames too to be visible - whereas on aarch64 the ABI keeps a frame pointer, so it works out of the box under `{$W-}`.

## Validation

11 assertions x 3 targets (FPC 3.2.3: x86_64-linux, aarch64-linux on RK3588, win64): non-empty text, 3-level nested call order, depth limiting, skip semantics, TTextWriter overload consistency.

### DIFF
--- a/src/core/mormot.core.log.pas
+++ b/src/core/mormot.core.log.pas
@@ -117,6 +117,15 @@ type
     diInternalMab,
     {$ifdef FPC} diInternalDwarf, diExternalDwarf {$else} diExternalMap {$endif});
 
+  /// how stack trace shall be computed during logging
+  // - stOnlyAPI is the first (and default) value, since manual stack makes
+  // unexpected detections, and was reported as very slow on Windows 11
+  // - on FPC, these values are ignored, because RTL CaptureBacktrace() is used
+  TSynLogStackTraceUse = (
+    stOnlyAPI,
+    stManualAndAPI,
+    stOnlyManual);
+
   /// allow to customize TDebugFile.Create and TDebugFile.SaveToFile process
   TDebugFileScope = set of (
     dfsIncludePathInFileName,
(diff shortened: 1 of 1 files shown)


src/core/mormot.core.log.pas
```diff
@@ -4062,6 +4078,71 @@ class function TDebugFile.AddLog(W: TTextWriter; aPointer: PtrUInt; NoHex: boole
...
+
+// capture the current thread stack into frames[], returning the frames count
+// - first frame is the caller of this function, plus optional skip levels
+function RawStackTrace(skip: PtrInt; var frames: TRawStackFrames): PtrInt;
+begin
+  {$ifdef FPC}
+  result := CaptureBacktrace(skip + 1, length(frames), pointer(@frames));
+  {$else}
+  {$ifdef OSWINDOWS}
+  result := RtlCaptureStackBackTrace(skip + 1, length(frames), @frames, nil);
```
synopse: there is also stOnlyManual mode on Windows 32-bit which should be supported.

synopse (changes requested):

src/core/mormot.core.log.pas
zen010101: Done - and your remark went further than the simple fallback I first wrote:

- new `use: TSynLogStackTraceUse` parameter (default stManualAndAPI), following the TSynLogFamily property semantics: ignored on FPC, stOnlyManual / stOnlyAPI honored on Delphi Windows, with the Win32 manual ebp-chain walk shared with AddStackManual (CheckAsmX86 is now a single shared function)
- the public path also sets tiExceptionIgnore as AddStackTrace does, so the speculative reads of the manual scan do not reach the intercepted-exception logs
- runtime-tested on Delphi 13 Win32: the first shared version of the scan was re-reading its own output frames[] buffer on stack (self-amplifying garbage frames) - fixed by skipping that range; stOnlyManual now locates the calling frame on a real Win32 build - deeper frames are not guaranteed, matching your "unexpected detections" comment on the enum
- validated on FPC 3.2.3 (x86_64-linux, aarch64-linux, win32, win64) and Delphi 13 (win32 incl. stOnlyManual at runtime, win64)

synopse:
Thanks for the input!

PR #560, opened 2026-08-31
core: rewrite POSIX RunRedirect() from popen() to fork/execve - fix #463

### LINKED ISSUE #463: RunRedirect on POSIX uses popen/sh -c, causing orphan child processes in TSynAngelize
Hi ab,

`RunRedirect` on POSIX (`mormot.core.os.posix.inc:4624`) uses FPC's `popen()` to launch processes:

```pascal
if popen(f, cmd, 'r') <> 0 then
  exit;
pid := pcint(@TFileRec(f).userdata[2])^;
```

`popen()` internally spawns `/bin/sh -c <cmd>`, which creates an extra shell layer in the process tree:

```
TSynAngelizeRunner thread
  └── popen() → /bin/sh -c "/path/to/worker --run"    ← pid recorded here
                  └── /path/to/worker --run             ← actual worker process
```

This causes two problems:

### 1. Orphan processes on stop

When TSynAngelize stops a `start:` service, `Runner.Abort` triggers `fpkill(pid, SIGTERM)` (line 4649). Since `pid` is the **shell process**, not the actual worker, the shell exits but the worker becomes an orphan process (reparented to PID 1) and keeps running. `TSynAngelize --kill` reports success, `--state` shows Stopped, but the worker is still alive.

### 2. Unnecessary shell overhead

`RunCommand` already has the logic to avoid the shell for simple commands via `ParseCommandArgs` + direct `FpExecve` (line 4562). But `RunRedirect` always goes through `popen()` regardless of command complexity.

### Suggested fix

Replace `popen()` with a direct `fork` + `FpExecve` + `pipe` approach, similar to what `RunCommand`/`RunInternal` already does. FPC's `TProcess` class also provides pipe redirection without the shell layer, for reference. The key points would be:

- Use `ParseCommandArgs` to split the command, then `fork` + `execve` directly (no `/bin/sh -c`)
- Create a pipe pair before fork for stdout capture (replacing `popen`'s implicit pipe)
- Record the actual child PID so that `fpkill(pid, SIGTERM)` reaches the right process

As a short-term alternative, changing line 4649 from `fpkill(pid, ...)` to `fpkill(-pid, ...)` (negative pid = process group) would at least ensure the worker gets the signal together with the shell.

zen010101:
Fixes #463 - plus a 2023 busy-spin regression found while writing the tests for it.

## 1. Orphan processes (#463)

`RunRedirect()` on FPC POSIX used `popen()`, i.e. always `/bin/sh -c`, recording the shell pid. On abort, `fpkill(shell)` leaves the actual worker re-parented to PID 1. Reproduced on Debian:

```
$ sh -c "sleep 9999; echo done" &   # sh pid=5289, sleep pid=5291
$ kill -TERM 5289
$ ps -o pid,ppid,args -p 5291
 5291       1 sleep 9999            # orphan
```

Now: `fork`/`pipe`/`execve`, and the child calls `setsid()` so the whole process group can be signaled - SIGTERM then SIGKILL to `-pid` kills the shell *and* its children. Simple commands are resolved in `$PATH` before fork and executed directly with no shell at all (with a shell fallback, e.g. for built-ins like `exit 3`), so in the common case no intermediate process even exists.

## 2. EOF busy-spin regression (since 7ec04e578, 2023/01)

The very first baseline test (`echo` + INFINITE) never returned: 7ec04e578 changed the pipe-read check from `n <= 0` to `n < 0`, so EOF (`read() = 0`) no longer terminates the loop. strace shows a tight `poll(POLLHUP)=1` / `read()=0` loop at 100% CPU. Since 2023/01, any `RunRedirect()` call with `waitfordelayms=INFINITE` hangs forever once the process ends, and calls with a finite timeout always burn the full timeout at 100% CPU. Fixed by restoring `n <= 0`.

## 3. Other behavior fixes

- `env` / `options` parameters are now honored (`popen()` silently ignored them, as the old comment stated);
- `wrkdir` now applies to the child process only - it used to `ChDir()` the whole calling process, and never restored it;
- `exitcode^` is now the actual `WaitProcess()` result, or -1 when aborted (it used to receive the raw `pclose()` wait status, e.g. 27647);
- the child environment is prepared before `fork()` (heap allocation after fork is not safe in a multi-threaded process), sharing the new `RunBuildEnv()` with `RunInternal()` - which also fixes `RunInternal()`'s E2BIG path that could `exit` inside the child and resume the caller's code in the forked process.

## Validation

22 assertions on aarch64-linux (RK3588 board) + x86_64-linux (FPC 3.2.3): output capture, `$PATH` resolution, shell pipes, timeout-kill orphan counts for simple and compound commands (the orphan probe positively self-checked via /proc scanning), env with/without `roEnvAddExisting`, wrkdir isolation, `onoutput` streaming + abort, built-in fallback exit codes. A/B against the `popen()` version: orphans 2 -> 0, garbage exitcode 27647 -> -1, busy-spin -> clean 607 ms return on a 500 ms timeout.

### DIFF
--- a/src/core/mormot.core.os.posix.inc
+++ b/src/core/mormot.core.os.posix.inc
@@ -5247,20 +5247,83 @@ begin
   end;
 end;
 
+type
+  // stack-allocated storage for RunBuildEnv() child environment pointers
+  TRunEnv = array[0..511] of PUtf8Char; // max 512 environment variables
+
+// compute the child environment pointers from Windows-layout env pairs
+// - to be called before fork(), since it may not be fork-safe
+// - returns false on too many environment variables
+// - returns true and envpp=nil to reuse the current process environment,
+// or envpp<>nil pointing to e[] entries (values from env and current envp)
+function RunBuildEnv(const env: TRunArg; options: TRunOptions;
+  var e: TRunEnv; out envpp: PPUtf8Char): boolean;
+var
+  cur: PPUtf8Char;
+  P: PUtf8Char;
+  n: PtrInt;
+begin
+  envpp := pointer(envp); // Delphi POSIX uses wrongly a PWideChar here
+  if env <> '' then
+  begin
+    result := false; // -ESysE2BIG on too many variables
+    n := 0;
+    if (roEnvAddExisting in options) and
+       (envpp <> nil) then
+    begin
+      cur := envpp;
+      while cur^ <> nil do
+      begin
+        if PosChar(cur^, #10) = nil then
+        begin
+          // filter to add only single-line variables
+          if n = high(e) - 1 then
+            exit;
+          e[n] := cur^;
+          inc(n);
+        end;
+        inc(cur);
+      end;
+    end;
+    P := pointer(env); // env follows Windows layout 'n1=v1'#0'n2=v2'#0#0
+    while P^ <> #0 do
+    begin
+      if n = high(e) - 1 then
+        exit;
+      e[n] := P; // makes POSIX compatible
+      inc(n);
+      inc(P, StrLen(P) + 1);
+    end;
+    e[n] := nil; // end with null
+    envpp := @e;
+  end;
+  result := true;
+end;
+
+function RunFork: TPID; {$ifdef HASINLINE} inline; {$endif}
+begin
+  // vfork is required with libc on BSD/SunOS - as FPC RTL POpen_internal()
+  {$if (defined(BSD) or defined(SUNOS)) and defined(FPC_USE_LIBC)}
+  result := FpvFork;
+  {$else}
+  result := FpFork;
+  {$ifend}
+end;
+
 function RunInternal(args: PPAnsiChar; waitfor: boolean; const env: TRunArg;
   options: TRunOptions): integer;
 var
   pid: TPID;
-  e: array[0..511] of PUtf8Char; // max 512 environment variables
+  e: TRunEnv;
   envpp: PPUtf8Char;
-  P: PUtf8Char;
-  n: PtrInt;
 begin
-  {$if (defined(BSD) or defined(SUNOS)) and defined(FPC_USE_LIBC)}
-  pid := FpvFork;
-  {$else}
-  pid := FpFork;
-  {$ifend}
+  // prepare the child environment before fork() for fork-safety
+  if not RunBuildEnv(env, options, e, envpp) then
+  begin
+    result := -ESysE2BIG;
+    exit;
+  end;
+  pid := RunFork;
   if pid < 0 then
   begin
     // fork failed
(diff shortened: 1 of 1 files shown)


synopse (changes requested):
Thanks for your subimission!

I think this needs a small adjustment before merge, mainly for BSD/libc compatibility.

`RunRedirect()` now unconditionally does:

```pascal
pid := FpFork;
```

whereas both FPC RTL `POpen_internal()` and the existing `RunInternal()` use `FpVFork` for:

```pascal
(BSD or SUNOS) and FPC_USE_LIBC
```

Could we keep the same behavior, ideally with a shared helper, e.g.:

```pascal
function RunFork: TPID; inline;
begin
  {$if (defined(BSD) or defined(SUNOS)) and defined(FPC_USE_LIBC)}
  result := FpVFork;
  {$else}
  result := FpFork;
  {$endif}
end;
```

and then use:

```pascal
pid := RunFork;
```

in both `RunInternal()` and `RunRedirect()`.

For the `vfork` case it would also be safer to keep the child side strictly to low-level calls and check the important failures:

```pascal
if pid = 0 then
begin
  if FpSetSid < 0 then
    FpExit(127);
  if FpDup2(fds[1], 1) < 0 then
    FpExit(127);
  FpClose(fds[0]);
  if fds[1] <> 1 then
    FpClose(fds[1]);
  if (wdir <> '') and
     (FpChdir(pointer(wdir)) <> 0) then
    FpExit(127);
  FpExecve(a[0], @a, envpp);
  FpExit(127);
end;
```

There is another small issue with PATH resolution. Currently:

```pascal
pathenv := RawUtf8(GetEnvironmentVariable('PATH'));
```

uses the parent's PATH, but `FpExecve()` receives `envpp`, which may contain a caller-supplied different PATH.

One possible fix is to get PATH from the already-built child environment:

```pascal
function RunEnvPath(envpp: PPUtf8Char): RawUtf8;
var
  p: PPUtf8Char;
begin
  result := '';
  p := envpp;
  if p = nil then
    p := pointer(envp);
  if p <> nil then
    while p^ <> nil do
    begin
      // keep the last PATH=, so explicit env values override inherited ones
      if IdemPChar(p^, 'PATH=') then
        FastSetString(result, p^ + 5, StrLen(p^ + 5));
      inc(p);
    end;
end;
```

then make `RunSearchPath()` accept the effective PATH:

```pascal
function RunSearchPath(name: PAnsiChar;
  const pathenv: RawUtf8): RawUtf8;
```

and reorder preparation slightly:

```pascal
if not RunBuildEnv(env, options, e, envpp) then
  exit;

childpath := RunEnvPath(envpp);

...

if err = [] then
begin
  path := RunSearchPath(a[0], childpath);
  ...
end;
```

This would make the new implementation closer to the existing/FPC BSD behavior and ensure that executable lookup uses the same environment that is eventually passed to `execve()`.

zen010101:
Thanks for the detailed review - all three points are applied as you sketched them:

- `RunFork()` shared helper (vfork with libc on BSD/SunOS, as `RunInternal()` and the RTL `POpen_internal()`), now used by both `RunInternal()` and `RunRedirect()`
- the child side is strictly limited to low-level syscalls, with `FpSetSid` / `FpDup2` failures checked and ending in `FpExit(127)`
- `RunEnvPath()` retrieves PATH= from the environment actually supplied to `execve()` (keeping the last PATH= so explicit env values override inherited ones), `RunSearchPath()` now takes that effective PATH, and `RunBuildEnv()` moved before the command preparation

Two new test assertions pin the PATH semantics down: a probe script reachable only through a caller-supplied child PATH is found and executed, and a command visible in the parent PATH (`ls`, not `echo` which is a shell built-in ignoring PATH) is NOT found once the child PATH excludes it - both green on x86_64-linux and aarch64-linux (FPC 3.2.3). The vfork branch itself compiles out on Linux, so it could not be exercised here - it mirrors the `RunInternal()` pattern verbatim.

synopse (approved):

PR #563, opened 2026-09-02
core: retain repeatedly reused small-block pools

Moonbot-Tech:
## Problem

When a sequential-feed pool contains one allocated block, freeing that block returns the whole pool to the medium allocator. A subsequent request for the same size recreates a pool under the shared lock. Repeated single-block churn therefore pays the full pool teardown/rebuild cost on every cycle.

## Change

For `FPCMM_SERVER`:

- the first eight empty-pool cycles still release the pool;
- from the ninth cycle onward, one current pool is retained for that size class;
- only physical block classes above 256 bytes participate, so per-thread tiny arenas are untouched;
- one byte of existing padding stores the counter, keeping `TSmallBlockType` at 64 bytes;
- normal `GetMem` gets no additional branch; the logic runs only in the already locked last-free path.

With the current size table, retained pool capacity is bounded at about 1.18 MiB process-wide, with no per-thread reservation.

## Measurements

A/B on current master (`aef05e96d`), Win64 x86-64, `-O3`, default `FPCMM_SERVER`:

| Case | Before | After | Speedup |
|---|---:|---:|---:|
| alloc/free 256 | 16.81 ns | 7.40 ns | 2.27x |
| alloc/free 512 | 16.83 ns | 7.42 ns | 2.27x |
| alloc/free 1024 | 17.18 ns | 7.40 ns | 2.32x |
| alloc/free 2048 | 16.82 ns | 7.41 ns | 2.27x |
| fragmented mixed | 107.86 ns | 10.31 ns | 10.46x |
| ring 512 | 3.827 ns | 3.822 ns | parity |

Under BOOSTER, the targeted churn cases improved by about 2.37x, while ring and mixed workloads remained around parity. Three alternating Horse server A/B rounds also remained around parity with a small positive median shift.

## Validation

- The GUI build is byte-for-byte identical before and after the change.
- GUI, SERVER and BOOSTER profiles build and run on Win64.
- Full MM qualification passed on Win64 SERVER/BOOSTER and Linux BOOSTER: 12 million randomized alloc/realloc/free operations, realloc matrices, cross-thread ownership transfers, pool lifecycle boundaries and 30 seconds of 26-thread lock saturation.
- The benchmark and ownership-safe MM stress programs are available in [MoonCompiler qualification](https://github.com/Moonbot-Tech/MoonCompiler/tree/main/qualification).

We found this optimization after compiling and load-testing the Brazilian [Horse](https://github.com/HashLoad/horse) web framework with [MoonCompiler](https://github.com/Moonbot-Tech/MoonCompiler).

### DIFF
--- a/src/core/mormot.core.fpcx64mm.pas
+++ b/src/core/mormot.core.fpcx64mm.pas
@@ -990,6 +990,7 @@ const
   SmallBlockDownsizeCheckAdder = 64;
   SmallBlockUpsizeAdder        = 32;
   SmallBlockTypePO2            = 6;  // SizeOf(TSmallBlockType)=64
+  SmallBlockPoolRetentionThreshold = 8;
 
   MediumBlockPoolSize          = MediumBlockPoolSizeMem - 16;
   {$ifdef FPCMM_MEDIUMPERTHREAD}
@@ -1071,7 +1072,8 @@ type
     GetmemCount: cardinal;
     FreememCount: cardinal;
     LastFreeLocked: boolean;
-    Padding: array[1 .. 3] of byte;
+    EmptyPoolReuseCount: byte;
+    Padding: array[1 .. 2] of byte;
     LastFreeCount: cardinal;
   end;
   PSmallBlockType = ^TSmallBlockType;
@@ -2548,6 +2550,7 @@ asm     // P = rcx on Windows, P = rdi on SystemV; use rsi = TSmallBlockType
         mov     rax, [rdx].TSmallBlockPoolHeader.FirstFreeBlock
         sub     [rdx].TSmallBlockPoolHeader.BlocksInUse, 1
         jz      @PoolIsNowEmpty
+@StoreFreeBlock:
         // Store this as the new first free block
         mov     [rdx].TSmallBlockPoolHeader.FirstFreeBlock, rcx
         // Store the previous first free block as the block header
@@ -2584,7 +2587,11 @@ asm     // P = rcx on Windows, P = rdi on SystemV; use rsi = TSmallBlockType
 @PoolIsNowEmpty:
         // FirstFreeBlock=nil means it is the sequential feed pool with a single block
         test    rax, rax
+        {$ifdef FPCMM_SERVER}
+        jz      @EmptySequentialFeedPool
+        {$else}
         jz      @IsSequentialFeedPool
+        {$endif FPCMM_SERVER}
         // Pool is now empty: Remove it from the linked list and free it
         mov     rax, [rdx].TSmallBlockPoolHeader.PreviousPartiallyFreePool
         mov     rcx, [rdx].TSmallBlockPoolHeader.NextPartiallyFreePool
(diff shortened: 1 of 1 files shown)


synopse:
It is weird this case happens enough it is even measurable. ;)

synopse:
I have now a big concern with this PR.

Once EmptyPoolReuseCount >= 8, a single-block sequential pool that becomes empty is sent to @StoreFreeBlock instead of being returned to the medium allocator. Since the counter never decays, that size class has effectively entered a permanent “retain single-block pool” mode.

I will investigate for a fix.

synopse:
I think `EmptyPoolReuseCount` should be reset when the sequential pool actually serves more than one block.

Currently the counter only increases, so after 8 single-block churn events the block type permanently enters retention mode, even if the workload later changes. In a long-running server this can unnecessarily keep small pools — and potentially their backing medium pools — alive.

Proposed reset:

```pascal
@TrySmallSequentialFeed:
        // Feed a small block sequentially
        movzx   ecx, [rsi].TSmallBlockType.BlockSize
        mov     rdx, [rsi].TSmallBlockType.CurrentSequentialFeedPool
        add     rcx, rax
        // Can another block fit?
        cmp     rax, [rsi].TSmallBlockType.MaxSequentialFeedBlockAddress
        ja      @AllocateSmallBlockPool
        {$ifdef FPCMM_SERVER}
        // More than one block fits: no longer consecutive single-block churn
        mov     byte ptr [rsi].TSmallBlockType.EmptyPoolReuseCount, 0
        {$endif FPCMM_SERVER}
        // Adjust number of used blocks and sequential feed pool
        mov     [rsi].TSmallBlockType.NextSequentialFeedBlockAddress, rcx
        add     [rdx].TSmallBlockPoolHeader.BlocksInUse, 1
```

This makes the threshold represent consecutive single-block churn:

* repeated 1-block allocate/free still benefits from pool retention;
* as soon as the pool needs a second block, the heuristic resets;
* when the workload later drains, pools can return through the normal medium-block coalescing/release path.

The performance impact should be negligible: one byte store on the already-locked/hot `TSmallBlockType` cache line, and it does not affect the actual one-block churn path once a retained free block is being reused.

This does not implement idle-time scavenging: if the application stops using a size class immediately after an empty pool has been retained, that last pool can still remain cached indefinitely. I think that is a separate issue from preventing stale lifetime history in this counter.

synopse:
Please try with my above commit.

synopse:
BTW when you benchmark our MM against FPC MM, I guess you use the one from latest trunk, i.e. the one rewritten by Rika?

Moonbot-Tech:
Thanks — your concern was correct. We had missed that lifetime edge case.

We tested your follow-up commit with our `MOONSHARD` profile. Resetting the counter on the sequential allocation path caused reproducible ring regressions of up to 8.9% (`1.0126×` geomean), so we kept your idea but moved the reset to the cold path where a multi-block sequential pool is fully drained. This leaves the `GetMem` path unchanged. The resulting A/B is `0.9998×`, and the new lifecycle regression passes under default, `FPCMM_SERVER`, `FPCMM_BOOSTER`, and the MoonCompiler product profile on both Win64 and Linux.

After your question, we also repeated the comparison against the exact current FPC trunk `heap.inc` (`da47439d`, including Rika’s rewrite), with the same compiler, RTL, test binary, and workload. Current Rika MM is close to our previous pinned FPC MM (`0.9942×` geomean). In the same 25-case allocator suite, the MoonCompiler profile built on your `fpcx64mm` runs in `0.4186×` the time of current Rika MM — about `2.39×` faster.

So yes: your `fpcx64mm` remains one of the main reasons our product runtime performs as it does. Thank you again for `fpcx64mm`, and for catching the lifetime issue here.

synopse:
thanks for the feedback.
so we better need to change something for my previous patch ? please make a PR.

I have also rewritten some part of the medium allocation path.
Mainly:
1) now medium block prefetch is also done outside of the lock - did not change anythi,ng
2) FPCMM_MEDIUMTOLARGE = on medium contention (with one spinning round), blocks > 48KB fallback to the Large allocator instead of sleeping
The latest patch was very simple, and in fact it reduces the medium contention a lot. the Large allocator is able to go back to medium allocation on resize.

I have enabled FPCMM_MEDIUMTOLARGE on FPCMM_SERVER and up, but feedback is welcome!
Perhaps we don't need it when FPCMM_MEDIUMPERTHREAD is defined... please try and report!

PR #566, opened 2026-09-02
Fix THttpMultiPartStream duplicated closing boundaries on keep-alive POST - fix #565

### LINKED ISSUE #565: THttpClientSocket: keep-alive connection corrupted after POST with THttpMultiPartStream, causing "400 Bad Request - Invalid Verb" on next request
I am not at all sure the problem isn´t sitting in front of my screen but here we go:

When sending two consecutive POST requests with the first containing a `THttpMultiPartStream` body on the **same** `THttpClientSocket` keep-alive connection, the **second** request fails on the server side with:

HTTP/1.1 400 Bad Request
Server: Microsoft-HTTPAPI/2.0
...
Bad Request - Invalid Verb
HTTP Error 400. The request verb is invalid.

....

Both requests are individually well-formed: uploading the file **alone** (fresh connection, no prior request on it) always succeeds.  The failure only occurs when the next POST Request immediately follows the upload on the same `THttpClientSocket` instance/connection.

**Workaround:** closing and reopening the `THttpClientSocket` connection between the two upload calls (i.e. never reusing the keep-alive connection across two `THttpMultiPartStream` POSTs) reliably eliminates the issue.

## Environment

- mORMot2 version/commit: 2.4.16411
- Delphi version and target: Delphi 12/32-bit Windows
- OS:  Windows Server 2022 with IIS / HTTP.sys backend (server, `Microsoft-HTTPAPI/2.0`)
- Transport: plain HTTP (port 80), `THttpClientSocket` with `KeepAlive > 0`
- Server-side authentication: Windows Integrated Authentication (Negotiate/NTLM via SSPI) 

## Trace of second (failing) Request

RequestInternal URL flags=pt idle=0 retry=0
mormot.net.client.THttpClientSocket(05f36440) DoRetry connection broken (socketerror=Ok) socket=1580 fatal=666 twice=false
mormot.net.client.THttpClientSocket(05f36440) open(URL) sock=1580 
mormot.net.client.THttpClientSocket(05f36440) RequestInternal URL flags= idle=0 retry=1
mormot.net.client.THttpClientSocket(05f36440) GetHeader=hrsNoStateMachine   flags=p len=31 application/json
mormot.net.client.THttpClientSocket(05f36440) GetBody len=31
mormot.net.client.THttpClientSocket(05f36440) RequestInternal status=401 flags=p in 1.89ms
mormot.net.client.THttpClientSocket(05f36440) Request(POST URL)=401
mormot.net.client.THttpClientSocket(05f36440) RequestInternal URL flags=p idle=0 retry=3
mormot.net.client.THttpClientSocket(05f36440) RequestInternal: response during SockSendStream {THttpMultiPartStream(090561b0):{}}
mormot.net.client.THttpClientSocket(05f36440) GetHeader=hrsNoStateMachine   flags=pc len=326 text/html; charset=us-ascii
mormot.net.client.THttpClientSocket(05f36440) GetBody len=326
mormot.net.client.THttpClientSocket(05f36440) RequestInternal status=400 flags=pc in 188us
mormot.net.client.THttpClientSocket(05f36440) DoSspi Authorization: Negotiate 400
RESPONSE BODY:  Bad Request - Invalid Verb HTTP Error 400. The request verb is invalid.

## Steps to reproduce

```pascal
var
  Http: THttpClientSocket;
  PostData: THttpMultiPartStream;
begin
  Http := THttpClientSocket.Open(Server, Port);
  Http.KeepAlive := 30000; // keep-alive enabled

  // --- Request A: small file (e.g. 8 KB XML) ---
  PostData := THttpMultiPartStream.Create;
  try
    PostData.AddContent('Payload', '[{"key":"...","orgFilename":"file.xml"}]', 'application/json');
    PostData.AddFileContent('~inStream', 'file.xml', XmlBytes, 'application/xml');
    PostData.Flush;
    Http.Post(Url, PostData, PostData.MultipartContentType, 10000); // -> 200 OK
  finally
    PostData.Free;
  end;

  // --- Request B: larger file (e.g. 110 KB PDF), SAME Http/connection ---
  PostData := THttpMultiPartStream.Create;
  try
    PostData.AddContent('Payload', '[{"key":"...","orgFilename":"file.pdf"}]', 'application/json');
    PostData.AddFileContent('~inStream', 'file.pdf', PdfBytes, 'application/pdf');
    PostData.Flush;
    Http.Post(Url, PostData, PostData.MultipartContentType, 10000);
    // -> FAILS: server responds 400 Bad Request - Invalid Verb (Microsoft-HTTPAPI/2.0)
    //    while the client is still streaming the body
    //    (trace log shows "response during SockSendStream")
  finally
    PostData.Free;
  end;

  Http.Free;
end;

landrix:
Fixes #565.

## Root cause

`THttpMultiPartStream.Flush` appends the closing `--boundary--` line on every call. The documented usage is to call `Flush` before `Post()`, and `THttpClientSocket.RequestInternal` then

1. computes `Content-Length:` from `InStream.Size` (in `CompressDataAndWriteHeaders`), and only afterwards
2. calls `InStream.Seek(0, soBeginning)` - which `TNestedStreamReader.Seek` turns into another `Flush`.

So the body actually streamed is 22 bytes longer than the header announces. The server consumes `Content-Length` bytes, answers, and then finds `--boundary--\r\n` on the socket: http.sys parses it as the next request line and answers `400 Bad Request - Invalid Verb` (the mORMot servers simply drop the connection). With NTLM/Negotiate the same stream instance is re-sent on every `RequestInternal` call of the handshake, and the overflow grows with each round trip - which matches the `response during SockSendStream` trace in the issue.

## Fix

- `THttpMultiPartStream.Flush` appends the closing boundaries only once (`fFlushed` flag), so it is safe to call it any number of times, e.g. on every retry. `Add*()` after `Flush` now raise `EHttpSocket` instead of silently producing a broken body (`AddFile` checks before opening the file, so nothing leaks).
- `THttpClientSocket.RequestInternal` rewinds `InStream` before `Content-Length:` is computed, so `Size` is right even when the caller never called `Flush` explicitly.

## Tests

- `TTestCoreBase.MultiPartDecoder`: `Flush`/`Seek(0)` keep `Size` unchanged, the stream reads back byte-identical, and `AddContent`/`AddFile` after `Flush` raise.
- `TNetworkProtocols.DoHttpBodyDownload`: the same `THttpMultiPartStream` is posted twice over one keep-alive connection, plus a never-flushed instance once, against both `THttpServer` and `THttpAsyncServer`; checks that exactly `Size` bytes were sent, `Size` did not grow, and the client never had to reconnect (`RequestContext` has no `DoRetry`).

Both tests were red before the fix (22 extra bytes per `Flush`; the async server variant even hung the second POST until the 10 s header timeout) and are green after it, on FPC 3.3.1 aarch64-win64 and Delphi 13 Win64.

### DIFF
--- a/src/net/mormot.net.client.pas
+++ b/src/net/mormot.net.client.pas
@@ -80,6 +80,8 @@ type
   // proper multipart formatting as defined by RFC 2488 / RFC 1341
   // - AddFile() won't load the file content into memory so it is more
   // efficient than MultiPartFormDataEncode() from mormot.core.buffers
+  // - THttpClientSocket.Post() will call Flush and rewind the stream itself,
+  // so the very same instance can be sent several times, e.g. on retry
   THttpMultiPartStream = class(TNestedStreamReader)
   protected
     fSections: THttpMultiPartStreamSections;
@@ -88,6 +90,7 @@ type
     fMultipartContentType: RawUtf8;
     fFilesCount: integer;
     fRfc2388NestedFiles: boolean;
+    fFlushed: boolean;
     function Add(const name, content, contenttype,
       filename, encoding: RawUtf8): PHttpMultiPartStreamSection;
   public
@@ -108,6 +111,9 @@ type
       const contenttype: RawUtf8 = '');
     /// call this method before any Read() call to sent data to HTTP server
     // - it is called also when Seek(0, soBeginning) is called
+    // - the closing boundaries are appended once, so it is safe to call this
+    // method several times, e.g. on every THttpClientSocket.Post() retry
+    // - no Add*() method should be called after Flush
     procedure Flush; override;
     /// the content-type header value for this multipart content
     // - equals '' if no section has been added
@@ -2293,6 +2299,8 @@ var
   ns: PtrInt;
   s: RawUtf8;
 begin
+  if fFlushed then
+    EHttpSocket.RaiseUtf8('%.Add(%) after Flush', [self, name]);
   // same logic than MultiPartFormDataEncode() from mormot.core.buffers
   ns := length(fSections);
   SetLength(fSections, ns + 1);
@@ -2384,6 +2392,8 @@ var
   fs: TStream;
   fn: RawUtf8;
 begin
+  if fFlushed then // check before opening the file, as Add() would do
+    EHttpSocket.RaiseUtf8('%.AddFile(%) after Flush', [self, filename]);
   fs := TFileStreamEx.Create(filename, fmOpenReadShared);
   // an exception is raised in above line if filename is incorrect
   StringToUtf8(ExtractFileName(filename), fn);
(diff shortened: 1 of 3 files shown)


synopse (approved):
Thanks for the feedback!

PR #569, opened 2026-09-02
core: skip medium-to-large fallback with per-thread arenas

Moonbot-Tech:
This follows up the `FPCMM_MEDIUMTOLARGE` question in #563.

`FPCMM_BOOSTER` already distributes medium allocations across four per-thread arenas and probes the other arenas before waiting. This change keeps `FPCMM_MEDIUMTOLARGE` enabled automatically for the ordinary `FPCMM_SERVER` profile, but stops enabling it automatically when `FPCMM_MEDIUMPERTHREAD` is active. Fine-grained users may still define `FPCMM_MEDIUMTOLARGE` explicitly.

Paired A/B on the same current source, 16 workers, sizes straddling the actual promotion boundary (48936/48937 bytes) and 49/64/100 KiB:

- Win64 BOOSTER, 9 alternating pairs: MEDIUMTOLARGE on/off = `1.017x` over promoted sizes;
- Linux x86-64 BOOSTER, 7 alternating pairs on CPUs 0-15: `1.006x`.

Once the four arenas are active, the fallback was parity to 1.7% slower in this mixed contention workload and gave no measured win. SERVER behavior is unchanged by this PR.

Validation of this profile change:

- Win64 and Linux builds report BOOSTER with `medpt` and without `medlrg`;
- SERVER still reports `medlrg` and without `medpt`;
- Linux `memory_mega quick`: PASS, including deterministic fuzz, cross-thread ownership, remote-free hot cases, and 26-thread saturation.

### DIFF
--- a/src/core/mormot.core.fpcx64mm.pas
+++ b/src/core/mormot.core.fpcx64mm.pas
@@ -60,7 +60,7 @@ unit mormot.core.fpcx64mm;
 
 // target a multi-threaded service on a modern CPU (default)
 // - define FPCMM_DEBUG, FPCMM_ASSUMEMULTITHREAD, FPCMM_ERMS, FPCMM_TINYPERTHREAD
-// and FPCMM_MEDIUMTOLARGE
+// and FPCMM_MEDIUMTOLARGE unless FPCMM_MEDIUMPERTHREAD is active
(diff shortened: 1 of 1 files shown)


synopse (approved):
Fair enough.
:)

PR #570, opened 2026-09-02
core: add opt-in FPCMM_MOONSHARD full small-block sharding

Moonbot-Tech:
Follow-up to https://github.com/synopse/mORMot2/pull/527#issuecomment-5512273641.

`FPCMM_BOOSTER` already shards tiny blocks up to 256 bytes and uses the four current user-medium arenas. Sizes from 272 through 2608 bytes still share one lock per size class. This opt-in profile extends the existing tiny-arena mechanism across all 44 small classes.

## Implementation

- 32 arenas selected by the existing thread-ID hash;
- the existing 44 size classes and the 2608-byte small/medium boundary stay unchanged;
- each arena row occupies 64 cache-line records (4096 bytes), so the hot-path arena selection remains shift/add;
- 20 physical tail records pad each row; no lookup selects them, and the first two in the primary row retain the existing cold fallback;
- the current medium allocator, including `FPCMM_MEDIUMPERTHREAD`, is untouched;
- `FPCMM_MOONSHARD` can be used alone or together with `FPCMM_BOOSTER`.

This is MoonCompiler's full-small arena topology rebuilt on top of present upstream. It deliberately keeps the current upstream GetMem and medium paths instead of restoring the older experimental 17KB small-block boundary or the superseded 32-medium-arena implementation.

## Validation

- Win64 O3: `SERVER`, `GUI`, and `BOOSTER` builds without `FPCMM_MOONSHARD` produce byte-identical MM objects and executables versus upstream `6340aef4a`.
- Win64 `memory_mega quick 424242`: PASS for `SERVER`, `MOONSHARD`, and `BOOSTER+MOONSHARD` (contracts, full size sweep, realloc matrix, pool lifecycle, cross-thread ownership, and 26-thread saturation).
- Linux FPC 3.2.2 `memory_mega quick 424242`: PASS for `SERVER`, `MOONSHARD`, `BOOSTER`, and `BOOSTER+MOONSHARD`.
- A 16-worker Win64 same-class A/B/B/A probe shows the intended contention ceiling. Throughput ranges in million alloc/free pairs per second:

| requested bytes | BOOSTER | BOOSTER + MOONSHARD |
| ---: | ---: | ---: |
| 320 | 6.43–7.20 | 288–296 |
| 800 | 7.16–7.44 | 287–347 |
| 1200 | 7.09–7.17 | 619–739 |
| 2600 | 6.45–7.36 | 633–664 |

The Linux `memspeed` small-allocation A/B/B/A gives the same trade-off on FPC 3.2.2: `THR-S x4` improves from 801 ms to 100 ms (8.0x), while the uncontended `ALLOC-S` pass moves from 37 ms to 58–59 ms. This profile is therefore intentionally opt-in for workloads where contention on 272–2608-byte classes matters; ordinary profiles remain byte-identical.

### DIFF
--- a/src/core/mormot.core.fpcx64mm.pas
+++ b/src/core/mormot.core.fpcx64mm.pas
@@ -55,7 +55,8 @@ unit mormot.core.fpcx64mm;
     1. default FPCMM_SERVER is perfect for a multi-threaded app/service/daemon;
     2. set FPCMM_GUI for GUI/console almost-mono-threaded apps (FastMM4 mode);
     3. try FPCMM_BOOST or FPCMM_BOOSTER on high-end hardware;
-    4. try mormot.core.fpclibcmm as POSIX alternative.
+    4. add FPCMM_MOONSHARD when small blocks > 256B are highly contended;
+    5. try mormot.core.fpclibcmm as POSIX alternative.
 }
 
 // target a multi-threaded service on a modern CPU (default)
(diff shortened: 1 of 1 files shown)


synopse (approved):
We best keep the FPCMM_MOONSHARD conditional naming in homage to your work and project, of course!

PR #575, opened 2026-09-07
A failed StartTransaction leaves the external storage locked for ever

dkounal:
When the DB connection dies at the moment a transaction is started, TransactionBegin() leaks state in two independent places. Both leaks are permanent: from that point on every TRestBatch send on that storage fails, until the process is restarted. 

Leak 1 — the session flag is never released
TRestStorageExternal.TransactionBegin (mormot.orm.sql.pas:1924):

if (aTable = fStoredClass) and
   inherited TransactionBegin(aTable, SessionID) then
  result := fProperties.SharedTransaction(SessionID, transBegin) <> nil
else
  result := false;

inherited is TRestOrm.TransactionBegin (mormot.orm.rest.pas:2376), which sets fTransactionActiveSession := SessionID and returns True. If SharedTransaction() then returns nil, the function returns False — but the session flag stays set. Every later call is refused at the inherited check, so this storage can never begin a transaction again.

Leak 2 — the entry is registered before the transaction is started
TSqlDBConnectionProperties.SharedTransaction (mormot.db.sql.pas:3724) registers the entry inside the lock (SessionID, RefCount := 1, Connection), and only afterwards, outside the lock, runs Connect / StartTransaction. If that raises, the handler sets result := nil and swallows the exception. The entry stays behind with RefCount = 1, describing a transaction that was never started.

This is worse than an error. With leak 1 fixed but leak 2 present, the next transBegin for the same SessionID takes the "found" branch and simply inc(t^.RefCount), returning success — so the caller writes outside any transaction, silently. Note the exit in that branch: a nested begin never reaches the SQL, so only the registering call can fail this way.


src/db/mormot.db.sql.pas
```diff
@@ -3806,13 +3812,49 @@ function TSqlDBConnectionProperties.SharedTransaction(SessionID: cardinal;
...
+        // OUTSIDE any transaction, silently
+        UnRegisterSharedTransaction(SessionID);
       result := nil; // result.StartTransaction/Commit/Rollback failed
       if action = transCommitWithException then
         raise;
     end;
   end;
 end;
 
+procedure TSqlDBConnectionProperties.UnRegisterSharedTransaction(
```
synopse, line 3831: Nice mormot-ish code!

synopse (approved):

PR #576, opened 2026-09-09
ldap - Add `TimeSpan` type storage with a function to make it readable

osvegn:
Some LDAP attributes are stored using TimeSpan format, such as PSO attributes. Some of those attributes are:
- `msDS-MaximumPasswordAge`
- `msDS-MinimumPasswordAge`
- `msDS-LockoutDuration`
- `msDS-LockoutObservationWindow`

ms-DS-Lockout-Duration attribute: https://learn.microsoft.com/en-us/windows/win32/adschema/a-msds-lockoutduration
TimeSpan Struct: https://learn.microsoft.com/en-us/dotnet/api/system.timespan?view=net-8.0

Does it makes sense to create a record to handle TimeSpan in mormot2?

### DIFF
--- a/src/net/mormot.net.ldap.pas
+++ b/src/net/mormot.net.ldap.pas
@@ -3946,7 +3946,7 @@ end;
 
 function TimeSpanToText(TimeSpan: Int64): RawUtf8;
 var
-  s, m, h, d: Int64;
+  s, m, h, d: Cardinal;
 begin
   s := Abs(TimeSpan) div 10000000;
   m := s div 60;
@@ -4018,8 +4018,7 @@ begin
       end;
     atsTimeSpan:
       begin
-        if TryStrToInt64(s, v) then
-          s := TimeSpanToText(v);
+        s := TimeSpanToText(GetInt64(@s));
       end;
     atsUnicodePwd:
       begin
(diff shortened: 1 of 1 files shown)


src/net/mormot.net.ldap.pas
```diff
@@ -3943,10 +3944,22 @@ procedure AttributeNameNormalize(var AttrName: RawUtf8);
     AttrName := RawUtf8(existing); // replace with existing interned name
 end;
 
+function TimeSpanToText(TimeSpan: Int64): RawUtf8;
+var
+  s, m, h, d: Int64;
```
synopse: cardinal for those variables seem enough

src/net/mormot.net.ldap.pas
```diff
@@ -4003,6 +4016,11 @@ function AttributeValueMakeReadable(var s: RawUtf8; ats: TLdapAttributeTypeStora
           exit;
         end;
       end;
+    atsTimeSpan:
+      begin
+        if TryStrToInt64(s, v) then
```
synopse: this is not Delphi/Unicode compatible.

synopse (changes requested):

synopse:
I will make some generic conversion routines in mormot.core.datetime.pas.

synopse:
beware: @s is not the same at all as pointer(s) ;)

I guess you would need some inverse text - to - timespan value too.

synopse:
@osvegn 
Are my current commits enough for OpenRSAT?

synopse:
I have added some text-to-timespan parsing functions:
https://github.com/synopse/mORMot2/commit/b6bef0425

@osvegn Hope it helps.

I am closing since on my side, all tests do pass.

PR #577, opened 2026-09-09
net: DHCP server was never sending the domain-name option 15

zen010101:
`TDhcpState.AddRegularOptions()` is documented twice (`mormot.net.dhcp.pas:1031` and `:6122`) as appending the `1,3,6,15,28,42` regular options, and the default `DHCP_REQUEST` client list does ask for option 15 - but `TDhcpScopeSettings.DomainName` was only ever used to build the DDNS FQDN, never written into the OFFER/ACK. So the documented `"domain-name"` setting silently had no effect, unless duplicated into a `"rule"` as `"always"`/`"requested"`.

The fix is one line, appended after `AddRulesOptions()` so a `"rule"` value still takes precedence, and using `AddOptionOnceU()` - as already used for options 66/67 - which skips an unset (nil) value, so scopes without a domain-name are unchanged.

Two cases added to `TNetworkProtocols.DHCP`: the scope value now shows up in the OFFER, and a `"rule"` one still overrides it. Checked both ways - with the line commented out again, the new assertion fails on the missing `domain-name`.

### DIFF
--- a/src/net/mormot.net.dhcp.pas
+++ b/src/net/mormot.net.dhcp.pas
@@ -4321,6 +4321,7 @@ begin
   AddOptionOnce32(doBroadcastAddress,   Scope^.Broadcast);
   AddOptionOnce32(doRouters,            Scope^.Gateway);
   AddOptionOnceA32(doDomainNameServers, pointer(Scope^.DnsServers));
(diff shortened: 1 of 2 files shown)


synopse (approved):
Sounds pretty legit.

Thanks for the feedback and test case.

PR #578, opened 2026-09-10
crypt: support original Keccak-256

Moonbot-Tech:
Adds original Keccak-256 for Ethereum-compatible hashing through the existing TSha3 sponge. SHA3-256 has different domain-separation padding and cannot substitute for it.

KECCAK_256 is appended to TSha3Algo without changing existing ordinal values. The patch adds its rate/padding selection and Keccak256Full / Keccak256 helpers; the permutation kernels are unchanged.

Validation: the existing TTestCoreCrypto._SHA3 suite passes 31,372 assertions in each Delphi 12.2 Win64 Release and Debug build, including both available x64 backends. New vectors cover empty input, abc, 135/136/137/272/1024-byte binary messages, every two-part split, and byte-at-a-time updates. Expected hashes were independently checked with PyCryptodome 3.23.0; all existing SHA-3 and SHAKE tests remain enabled. FPC and other targets were not run for this PR.

### DIFF
--- a/src/crypt/mormot.crypt.core.pas
+++ b/src/crypt/mormot.crypt.core.pas
@@ -2278,16 +2278,18 @@ type
   PSha512 = ^TSha512;
 
 type
-  /// SHA-3 instances, as defined by NIST Standard for Keccak sponge construction
+  /// SHA-3, SHAKE and original Keccak instances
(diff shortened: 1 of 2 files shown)


synopse (approved):
Sounds good.
Thanks!

PR #579, opened 2026-09-10
core: fix numeric conversion overflow and malformed decimal input

Moonbot-Tech:
Numeric conversions can silently wrap at signed integer and Currency boundaries. The patch validates mantissa/exponent ranges and decimal grammar, with Double fallback when enabled. It retains the current upstream NaN/Inf handling and normal conversion path.

Regression checks are included in the existing TTestCoreBase.NumericalConversions method and cover signed bounds, Currency limits, exponent overflow and malformed decimal input.

Validation on Win64, with a full rebuild for each configuration:
- Delphi 12.2: Release (O+) and Debug (O-, DEBUG).
- Stock FPC 3.2.2: Release (O3) and Debug (O-, DEBUG).
- All four runs pass NumericalConversions, Curr64 and Variants: over 2.68 million assertions per run, zero failures.
- A 40,000-number differential corpus on Delphi: GetExtended results match bit-for-bit; JSON results differ only for 188 previously overflowing Currency values, which now become Double.

### DIFF
--- a/src/core/mormot.core.text.pas
+++ b/src/core/mormot.core.text.pas
@@ -6726,6 +6726,10 @@ end;
 {$ifndef WIN32DELPHI} // Delphi has its own x86/x87 asm version
 
 function GetExtended(P: PUtf8Char; out err: integer): TSynExtended;
+const
+  Scale: double = 1.3407807929942597e154; // 2^512
+  InvScale: double = 7.458340731200207e-155; // 2^-512
+  MaxScaled: double = 1.3407807929942596e154; // MaxDouble * 2^-512
 var
   remdigit: integer;
   frac, exp: PtrInt;
@@ -6774,7 +6778,8 @@ begin
     if byte(ord(P^) - ord('0')) <= 9 then
     begin
       if (remdigit <> 0) or // avoid 64-bit overflow, but allow 19 digits
-         (v64 > 922337203685477580) then
+         (v64 > 922337203685477580) or
+         ((v64 = 922337203685477580) and (P^ > '7')) then
         dec(remdigit);
       if remdigit >= 0 then // over-required digits are just ignored
       begin
@@ -6792,13 +6797,15 @@ begin
     if P^ <> '.' then
       break;
     inc(P);
-    if frac > 0 then
+    if frac <> 0 then
       goto e; // will return partial value but err=1
     dec(frac);
   until false;
   inc(frac, ord(frac < 0)); // adjust digits after '.'
   if ord(P^) or $20 = ord('e') then
   begin
+    if not (fValid in flags) then
+      goto e;
     exp := 0;
     exclude(flags, fValid);
     inc(P);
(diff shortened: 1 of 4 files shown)


synopse:
I agree with the number conversion fixes.
Why not tested with FPC?

But the TTextWriter fix is IMHO overkill.
The getmem() of the internal is only up to 1MB so I don't see it would kill all memory in practice.
There will be plenty of other places which could need fixing first if we got short of memory.
Also the tests are clearly over-engereed for almost no benefit.

It is always better to make one PR per feature, not mix content like this.

src/core/mormot.core.text.pas
```diff
@@ -4707,17 +4708,19 @@ procedure TTextWriter.FlushToStream;
...
     // tune small (stack-allocated?) buffer to grow by twice its size
-    fTempBufSize := fTempBufSize * 2
+    tmp := tmp * 2
   else if (written > 40 shl 20) and
           (tmp < 1 shl 20) then
-    fTempBufSize := 1 shl 20 // total > 40MB -> grow internal buffer once to 1MB
+    tmp := 1 shl 20 // total > 40MB -> grow internal buffer once to 1MB
   else
     exit; // nothing to change about internal buffer size
+  GetMem(newbuf, tmp); // preserve buffer ownership and size if allocation fails
```
synopse: A 1MB allocation failure is perhaps our biggest concern...

test/test.core.base.pas
```diff
@@ -5872,6 +5878,225 @@ procedure TTestCoreBase.NumericalConversions;
   NotifyTestSpeed('StrInt64', 100000, 0, @Timer);
 end;
 
+procedure TTestCoreBase.NumericParsingBounds;
```
synopse: Should be included in an existing method.

test/test.core.base.pas
```diff
@@ -5872,6 +5878,225 @@ procedure TTestCoreBase.NumericalConversions;
...
+  CheckEqual(err, 0);
+  Check(not IsInfinite(d) and (d > 1.79e308));
+  s := '99e307';
+  d := GetExtended(pointer(s), err);
+  Check(err <> 0);
+  Check(not IsInfinite(d));
+  {$endif WIN32DELPHI}
+end;
+
+{$ifdef ISDELPHI2006ANDUP}
```
synopse: Clearly overkill for a no situation.

synopse (changes requested):

Moonbot-Tech:
Thanks for the review. Updated in d48f463:

- Removed the TTextWriter change and its allocation-failure tests from this PR.
- Moved the numeric regression checks into the existing NumericalConversions method.
- Tested with stock FPC 3.2.2 Win64, both Release (-O3) and Debug (-O-, DEBUG), and reran Delphi 12.2 Win64 Release and Debug. NumericalConversions, Curr64 and Variants pass in all four configurations, with over 2.68 million assertions per run and no failures.

One clarification about TTextWriter: the issue is the state left behind if GetMem raises. FlushToStream has already freed the old heap buffer, or cleared the external-buffer flag, so subsequent destruction can free the old buffer again or attempt to free an external buffer. The change was intended to preserve ownership on allocation failure, not to reduce memory consumption. I agree it should be separate from the numeric fixes, so it is no longer included here.

synopse (approved):

synopse:
@Moonbot-Tech 
I guess we got something fine now.
The 0.00000000000000000001 issue I just discovered was in fact really concerning, especially because Delphi Win32 properly support it via its FP80 asm.

Moonbot-Tech:
We checked, and there seems to be a bug with zero values: JSON `0.0` is rejected as invalid, and `GetExtended('0.0e400', err)` reports a range error, although both values are simply zero.

synopse:
I continued working on it today.
Please retry.
:)

Moonbot-Tech:
The zero cases work now, thanks! We found another edge case in `GetExtended`: the NaN/Inf check reads four bytes even when the string is shorter. For example, `N` plus its null terminator occupies two bytes. If the buffer ends at the boundary of an accessible page, this causes an access violation instead of a conversion error.

We added a check before that read. The fix is in a separate commit in the PR, where we also propose our ASM number parsing for Delphi Win64: https://github.com/synopse/mORMot2/pull/582.

PR #580, opened 2026-09-12
net: fixed DHCPNAK frames not following RFC 2131 Table 3

zen010101:
A `DHCPNAK` is built by `DoError(dsmNak)` and then goes through the shared `Flush()`, so it currently carries everything a `DHCPACK` does. Four points of RFC 2131 are not met:

- **`siaddr`** — Table 3 has it as `0` in a DHCPNAK, but `DhcpNew()` fills it with the server identifier for every message type.
- **options** — Table 3 excludes the lease time (51/58/59), the network settings (1/3/6/15/28/42) and the boot options from a DHCPNAK, but the shared `Flush()` appends them, plus any matching `"rule"` options. The option 61/82 copies done by `TDhcpState.Flush()` stay, as RFC 6842/3046 expect.
- **`chaddr`** — Table 3 wants the `chaddr` of the client DHCPREQUEST, but the NAK is built from `State.Mac64`, which `DhcpParse()` takes from option 61 when the client supplies one.
- **destination** — with no relay agent, 4.3.2 has *"the server MUST broadcast the DHCPNAK message to the 0xffffffff broadcast address because the client may not have a correct network address or subnet mask"*, but it goes to the scope broadcast address (e.g. `192.168.0.255`), which a client whose address is no longer in that subnet may well drop.

The relay case is unchanged: a NAK with `giaddr` set still goes to the relay agent, and `DhcpNew()` already sets the broadcast flag that 4.3.2 requires there. Moving the NAK case first in `OnFrameReceived` also lets the `dmtNak` condition be removed from the unicast branch.

### Tests

Three cases on the resulting frame, checked as JSON through `SendToJson`: a plain NAK, a NAK while a default `"rule"` matches, and a NAK from a client whose option 61 holds another MAC and which supplies an option 82 (so the same assertion covers the `chaddr` echo and the 61/82 copies). They are appended at the end of the DHCP test so the existing metrics assertions are untouched.

The broadcast destination lives in `TDhcpServerThread.OnFrameReceived`, which the `ComputeResponse()` level tests cannot reach — that one line is not covered.

`TNetworkProtocols` is 175,112 assertions passed (FPC 3.2.3, i386-win32).

### DIFF
--- a/src/net/mormot.net.dhcp.pas
+++ b/src/net/mormot.net.dhcp.pas
@@ -1893,7 +1893,8 @@ begin
   DhcpAddOptionByte(result, {doMessageType=} 53, ord(dmt));
   if serverid <> 0 then
   begin
-    dhcp.siaddr := serverid;
+    if dmt <> dmtNak then
+      dhcp.siaddr := serverid; // RFC 2131 Table 3: 'siaddr' is 0 in a DHCPNAK
     DhcpAddOption32(result, doServerIdentifier, serverid);
   end;
   result^ := #255;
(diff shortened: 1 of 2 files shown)


synopse:
Sounds like a nice fix.

PR #581, opened 2026-09-12
net: NAK a DHCPREQUEST conflicting with a static reservation

zen010101:
When a MAC already has a lease, the `dmtRequest` branch acknowledges the IP the server has, whatever the client did ask for. For dynamic and PXE clients that is deliberate — 39ba4b2ec used to say so in a comment: *"especially for PXE networks, it is safe and common to ignore Option 50 and ciaddr in REQUEST and ACK the already OFFERed IP"* — and **this PR leaves that path alone**.

A static reservation is a different statement though: the address is the one the administrator configured for that MAC. When the client insists on another one, ACKing a different `yiaddr` leaves it to the client whether it switches at all, and when. RFC 2131 4.3.2 has *"Server SHOULD send a DHCPNAK message to the client if the 'requested IP address' is incorrect"* on INIT-REBOOT, and *"The DHCP server SHOULD check 'ciaddr' for correctness before replying"* on REBINDING — a NAK makes the client DISCOVER and get the reserved address at once, which is what a standard server does here.

A NAK is returned only when all of these hold:

1. the lease is `lsStatic` — nothing from a dynamic pool is affected;
2. the client did ask for an explicit address, from option 50 (SELECTING / INIT-REBOOT) or from `ciaddr` (RENEWING / REBINDING) — a REQUEST asking for no address at all, as PXE firmware often sends, is still acknowledged;
3. that address is not the reserved one;
4. the REQUEST did not select another server through option 54 — RFC 2131 3.1 has *"Those servers not selected by the DHCPREQUEST message use the message as notification that the client has declined that server's offer"*.

### What this really changes

To be explicit, because it is the part worth your judgement: **this is a policy change, not merely the detection of an edited reservation.** The server has no memory of what the reservation used to be, so it cannot tell a stale client address from one the administrator just changed. A client of a never-modified reservation that insists on some other address now gets a NAK where it used to get an ACK.

That also reaches the `lsStatic` leases `FakeLease()` builds from an option 61/97 UUID or from a `"rule"` holding an `"ip"` — for instance when a PXE chainload presents a different identity than the firmware stage did, and keeps asking for the address of the previous stage.

If that tolerance is something you want to keep for static entries too, this PR is the wrong shape and I would rather drop it than weaken the reasoning.

`DhcpIP4()` had to be fixed first: it returned the option offset instead of `0` when the option is present with a length other than 4, which would make a malformed option 50 look like a conflicting address. No caller relies on that offset — both `GetScope(DhcpIP4(...))` ones already fall back when no scope matches.

### Tests

The two NAK paths (option 50 and `ciaddr`), plus everything that must *not* change: a dynamic lease is still acknowledged whatever is requested, the reserved IP and a matching `ciaddr` are acknowledged, a REQUEST with no requested IP at all is acknowledged, another server's REQUEST is left alone, unparsable option 50 or 54 values are ignored, and the DISCOVER following a NAK returns the reservation.

They are appended at the end of the DHCP test so the existing metrics assertions are untouched. `TNetworkProtocols` is 175,134 assertions passed (FPC 3.2.3, i386-win32).

This is independent of the DHCPNAK frame fixes in the other PR, and will be rebased if that one lands first — they do touch the same spot in the test file.

### DIFF
--- a/src/net/mormot.net.dhcp.pas
+++ b/src/net/mormot.net.dhcp.pas
@@ -1073,6 +1073,7 @@ type
     function NotifyDnsScript(opt81: PAnsiChar; len: PtrInt; plain: boolean): boolean;
     procedure AddOption81;
     procedure ParseRecvLensRai;
+    function RequestConflictsWith(ip4: TNetIP4): boolean;
     function FakeLease(found: TNetIP4): PDhcpLease;
       {$ifdef HASINLINE} inline; {$endif}
     function ClientUuid(opt: TDhcpOption): PDhcpLease;
@@ -1936,7 +1937,7 @@ end;
 
 function DhcpIP4(dhcp: PDhcpPacket; len: PtrUInt): TNetIP4;
 begin
-  result := len;
+  result := 0; // an option with an unexpected length is no IPv4 address
   if len = 0 then
     exit;
   len := PtrUInt(@dhcp.options[len]);
@@ -4378,6 +4379,25 @@ begin
   until len = 0;
 end;
 
+function TDhcpState.RequestConflictsWith(ip4: TNetIP4): boolean;
+var
+  req: TNetIP4;
+  len: PtrUInt;
+begin
+  result := false;
+  // RFC 2131 3.1: a REQUEST selecting another server is none of our business
+  len := RecvLens[doServerIdentifier];
+  if (len <> 0) and
+     (DhcpIP4(@Recv, len) <> Scope^.ServerIdentifier) then
+    exit;
+  // option 50 on SELECTING/INIT-REBOOT, or ciaddr on RENEWING/REBINDING
+  req := DhcpIP4(@Recv, RecvLens[doRequestedAddress]);
+  if req = 0 then
+    req := Recv.ciaddr;
+  result := (req <> 0) and
+            (req <> ip4);
+end;
+
 function TDhcpState.FakeLease(found: TNetIP4): PDhcpLease;
 begin
   result := @Temp; // fake transient PDhcpLease for this StaticUuid[]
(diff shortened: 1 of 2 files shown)


synopse:
Thanks for this PR. I’ve merged #580 first.

For #581, the fix itself seems legitimate, but please adjust two points before merging, about testing:

* The test labeled INIT-REBOOT currently includes option 54 (Server Identifier), which makes it a SELECTING request. Please add a real INIT-REBOOT case with option 50, `ciaddr = 0`, and no option 54.
* A DHCPREQUEST selecting another server should not be ACKed by us. The current test explicitly expects an ACK; this should instead result in no response.

Or did I miss something?

src/net/mormot.net.dhcp.pas
```diff
@@ -4378,6 +4379,25 @@ procedure TDhcpState.ParseRecvLensRai;
   until len = 0;
 end;
 
+function TDhcpState.RequestConflictsWith(ip4: TNetIP4): boolean;
+var
```
synopse, line 4383:
```
function TDhcpState.RequestConflictsWith(ip4: TNetIP4): boolean;
var
  ip: TNetIP4;
begin
  result := false;
  // RFC 2131 3.1: a REQUEST selecting another server is none of our business
  ip := DhcpIP4(@recv, RecvLens[doServerIdentifier]);
  if (ip <> 0) and
     (ip <> Scope^.ServerIdentifier) then
    exit;
  // option 50 on SELECTING/INIT-REBOOT, or ciaddr on RENEWING/REBINDING
  ip := DhcpIP4(@Recv, RecvLens[doRequestedAddress]);
  if ip = 0 then
    ip := Recv.ciaddr;
  if  (ip <> 0) and
      (ip <> ip4) then
    result := true;
end;
```

synopse (changes requested):

synopse:
**Update:**

I have tried to fix it more broadly.
See my commits above.

Thanks anyway for the feedback!

synopse:
I will close this PR.
Please re-open or create another PR if I missed something.

Thanks for the feedback and analysis!

zen010101:
Thanks for taking it further — the broader fix covers our case.

We run this on Debian 11 / NetworkManager 1.30.6, and your choice to leave RENEWING alone turns out to be the right one for that client. Measured today on real hardware (120s lease, reservation moved mid-lease):

- the T1 unicast renewal does get the NAK and **ignores it** — the node keeps the old address until T2, so covering `ciaddr` would have changed nothing here;
- after T2 it retries DISCOVER but rejects every OFFER that is not its previous address (`selecting lease failed: -131`, 6 times over 79s, no address on the interface meanwhile);
- it only recovers when it sends a REQUEST for the old address and gets your NAK: `canceled DHCP transaction` → new transaction → bound to the new reservation. 2m44s in total from moving the reservation.

So the option 50 path you implemented is not merely the compliant one, it is the one that actually unblocks this client. Nothing else to report from our side.

synopse:
To be fair, I did not finalize the DHCP server itself.
So you used it and find out its unfinished state the dirty way.
Sorry!

Don't hesitate to come back and give your feedback.
It is always very insightful.
Thanks again.

PR #582, opened 2026-09-12
Faster number parsing for Delphi Win64

Moonbot-Tech:
We use mORMot extensively to process exchange JSON and wanted to make number conversion faster. We brought our ASM routines over to the current upstream and adapted them to its behavior, including Currency, long mantissas, and rounding. The public API stays the same.

On our corpus of 780 REST/WS exchange responses, converting numbers stored as strings takes **42.5% less time**, and converting JSON numbers takes **39.1% less time**. This is Delphi 12.2 Win64 on a Ryzen 7 5800X. The difference for the full JSON processing cycle is more modest: from -2.19% to +0.38%.

We ran the existing tests, millions of comparisons against Pascal, real exchange responses, and separate checks for memory boundaries, register preservation, FPU behavior, and multithreading. Values, types, and errors matched in the tested cases; Double results matched bit for bit. We checked both Release and Debug, as well as the existing path under FPC Win64.

Strings with leading spaces or a plus sign, NaN/Inf, and malformed suffixes still fall back to Pascal to preserve its behavior. There is a small cost here: an extra 0.7-2.6 ns in our measurements. Double and Int64 boundary cases are handled directly in ASM.

The tests also caught a read past the end of short strings in the NaN/Inf check. That fix is in a separate commit. Regression tests were added to the existing `NumericalConversions`.

### DIFF
--- a/src/core/mormot.core.text.asmx64.inc
+++ b/src/core/mormot.core.text.asmx64.inc
@@ -1,0 +1,241 @@
+{ Delphi Win64 decimal conversion. Byte loads and volatile registers only.
+  Eighteen digits fit without overflow checks; further digits are accumulated
+  while the signed mantissa fits, following the portable precision rules.
+  Special syntax, partial error results and out-of-range scales use the portable
+  implementation. No floating-point operation precedes a fallback. }
+const
+  NumberSignMask: UInt64 = $8000000000000000;
+  NumberScale: double = 1.3407807929942597e154; // 2^512
+  NumberInvScale: double = 7.458340731200207e-155; // 2^-512
+  NumberMaxScaled: double = 1.3407807929942596e154; // MaxDouble * 2^-512
+
+function GetExtended(P: PUtf8Char; out err: integer): TSynExtended;
+asm .noframe
+        movq    xmm2, rcx // original input for the general path
+        test    rcx, rcx
+        jz      @slow
+        xor     r11d, r11d
+        cmp     byte ptr [rcx], '-'
+        jne     @positive
+        inc     rcx
+        inc     r11d
+@positive:
+        movzx   eax, byte ptr [rcx]
+        sub     eax, 48
+        cmp     eax, 9
+        ja      @slow
+        lea     r8, [rcx + 18] // up to 18 digits need no overflow check
+        xor     r9d, r9d
+        .align 16
+@integer:
+        cmp     rcx, r8
+        je      @moreInteger
+@integerDigit:
+        lea     r9, [r9 + r9 * 4]
+        lea     r9, [rax + r9 * 2]
+        inc     rcx
+        movzx   eax, byte ptr [rcx]
+        sub     eax, 48
+        cmp     eax, 9
+        jbe     @integer
+        mov     r10, rcx
+        cmp     eax, -2 // '.' minus '0'
+        jne     @finish
+        inc     rcx
+        mov     r10, rcx
+        movzx   eax, byte ptr [rcx]
+        sub     eax, 48
+        cmp     eax, 9
+        ja      @finish
+        .align 16
+@fraction:
+        cmp     rcx, r8 // last permitted byte; two digits fit while rcx < r8
+        jae     @lastFraction
+        lea     r9, [r9 + r9 * 4]
+        lea     r9, [rax + r9 * 2]
+        inc     rcx
+        movzx   eax, byte ptr [rcx]
+        sub     eax, 48
+        cmp     eax, 9
+        ja      @finish
+@fractionOne:
+        lea     r9, [r9 + r9 * 4]
+        lea     r9, [rax + r9 * 2]
+        inc     rcx
+        movzx   eax, byte ptr [rcx]
+        sub     eax, 48
+        cmp     eax, 9
+        jbe     @fraction
+@finish:
+        sub     r10, rcx
+        cmp     eax, -48
+        jne     @exponent
+        jmp     @checkScale
+@value:
+        lea     r8, [rip + POW10]
+        movsd   xmm0, qword ptr [r8 + r10 * 8 + 31 * 8] // POW10[frac]
+        pxor    xmm1, xmm1
+        cvtsi2sd xmm1, r9
+@multiply:
+        mulsd   xmm0, xmm1
+@applySign:
+        test    r11b, 1
+        jz      @store
+        movq    xmm3, qword ptr [rip + NumberSignMask]
+        xorpd   xmm0, xmm3
+@store:
+        mov     dword ptr [rdx], 0
+        ret
+@lastFraction:
+        ja      @moreFraction
+        jmp     @fractionOne
+@moreInteger:
+        movq    xmm4, rax
+        mov     rax, 922337203685477580
+        cmp     r9, rax
+        movq    rax, xmm4
+        ja      @discardInteger
+        jb      @retainInteger
+        cmp     eax, 7
+        ja      @discardInteger
+@retainInteger:
+        inc     r8
+        jmp     @integerDigit
+@discardInteger:
+        xor     r10d, r10d
+@discardIntegerLoop:
+        inc     r10
+        inc     rcx
+        movzx   eax, byte ptr [rcx]
+        sub     eax, 48
+        cmp     eax, 9
+        jbe     @discardIntegerLoop
+        cmp     eax, -48
+        je      @checkScale
+        jmp     @exponent
+@moreFraction:
+        movq    xmm4, rax
+        mov     rax, 922337203685477580
+        cmp     r9, rax
+        movq    rax, xmm4
+        ja      @discardFraction
+        jb      @retainFraction
+        cmp     eax, 7
+        ja      @discardFraction
+@retainFraction:
+        inc     r8
+        jmp     @fractionOne
+@discardFraction:
+        sub     r10, rcx
+@discardFractionLoop:
+        inc     rcx
+        movzx   eax, byte ptr [rcx]
+        sub     eax, 48
+        cmp     eax, 9
+        jbe     @discardFractionLoop
+        cmp     eax, -48
+        je      @checkScale
+        jmp     @exponent
+@exponent:
+        or      eax, 32
+        cmp     eax, 53 // 'e' or 'E', after subtracting '0' and setting bit 5
+        jne     @slow
+        inc     rcx
+        movzx   eax, byte ptr [rcx]
+        cmp     eax, '+'
+        je      @exponentSign
+        cmp     eax, '-'
+        jne     @exponentFirst
+        or      r11d, 2
+@exponentSign:
+        inc     rcx
+        movzx   eax, byte ptr [rcx]
+@exponentFirst:
+        sub     eax, 48
+        cmp     eax, 9
+        ja      @slow
+        mov     r8d, eax
+        jmp     @exponentNext
+@exponentLoop:
+        cmp     r8d, 1000000 // the portable path validates huge exponents
+        ja      @slow
+        lea     r8, [r8 + r8 * 4]
+        lea     r8, [rax + r8 * 2]
+@exponentNext:
+        inc     rcx
+        movzx   eax, byte ptr [rcx]
+        sub     eax, 48
+        cmp     eax, 9
+        jbe     @exponentLoop
+        cmp     eax, -48
+        jne     @slow
+        test    r11b, 2
+        jz      @exponentPositive
+        sub     r10, r8
+        jmp     @checkScale
+@exponentPositive:
+        add     r10, r8
+@checkScale:
+        lea     rax, [r10 + 31]
+        cmp     rax, 62
+        jbe     @value
+        test    r9, r9
+        jnz     @largeScale
+        xor     r10d, r10d
+        jmp     @value
+@largeScale:
+        cmp     r10, -342
+        jl      @slow
+        cmp     r10, 308
+        jg      @slow
+        pxor    xmm1, xmm1
+        cvtsi2sd xmm1, r9
+        lea     r8, [rip + POW10]
+        test    r10, r10
+        js      @negativeScale
+        cmp     r10, 290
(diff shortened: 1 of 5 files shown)


synopse:
Amazing.

But perhaps it would be possible to:
- include https://github.com/synopse/mORMot2/pull/583 logic fix
- include NAN/INF parsing in asm (both asm and pascal should have the same exact behavior)
- make a FPC compatible version (Linux + Win64)
- keep the asm in mormot.core.text.pas - not in a new .inc file yet; or move it into mormot.core.base.asmx64.inc just like GetExtended() for Win32 Delphi is already there - we would need to move POW[] constant to this unit; but keep the pure pascal version in mormot.core.text.

synopse (changes requested):
Please see my comment.

synopse:
To be honnest, I am not convinced by the actual need of the full 10 KB Eisel-Lemire table inclusion.
Sounds like really premature optimization to me.
I guess we could optimize further first the pure pascal path before rewritting everything in asm. 
Especially if there is no sensitive benefit with actual JSON parsing.

synopse:
I have rewritten the GetExtended() hot path in pure pascal.
Please try again with my updated  pure pascal code.
https://github.com/synopse/mORMot2/commit/71129606a
https://github.com/synopse/mORMot2/commit/c88c4ad7b
In practice, the new direct path for plain numbers, and Clinger's fast path with `result := result / POW10[-frac]` are likely to be faster than the Lemire code (which is fine with intrinsics and modern C compilers but much less with FPC), and much less error prone.
The nice benefit is that we could expect good compilation on aarch64 too.

Moonbot-Tech:
We tested your updated Pascal code. We also reworked the ASM path using SIMD without a large table and addressed your NaN/Inf and FPC comments.

On our exchange corpus, direct `GetExtended` conversion takes 5.6–6.0 ns per number versus 13.0–13.6 ns with Pascal. In-place parsing of large order books, including reading all bid/ask prices and quantities, is about 7–13% faster.

We also found a couple of small fixes for Pascal: trailing zeros could sometimes change the last bit of a Double, and the fractional scale was not adjusted correctly at the negative mantissa boundary. The fixes and regression tests are included.

synopse:
I merged it, since it sounded fine.
I guess we could require a SSE3 enabled CPU for Win64 and Linux x64.

I have just made a specific branch for FP80 / extended / x87 32-bit targets in the pascal code.

But I will need to slightly refactor your asm for proper cross-platform support with FPC.

synopse:
@Moonbot-Tech 
Thanks again for the work on this PR.

After integrating and refactoring the code, we extended our numerical benchmark with a broader mix of inputs: short integers, fixed-point values of different lengths, padded fractions, long fractions, and scientific notation.

We included some fixes, and a faster path for 1-2 integer digits.

But the results, from a pre-computed corpus of 60,000 values in `BenchmarkNumbers` function, show a fairly clear difference between `GetExtended()` and `GetNumericVariantFromJson()`.

For `GetExtended()`, the x64 SSSE3 implementation is consistently the fastest one on this mixed corpus.

Delphi Win64:

* SSSE3 ASM: 564 µs for 60k values
* scalar ASM: 614 µs
* pure Pascal: 687 µs

FPC/Linux x64:

* SSSE3 ASM: 487 µs
* pure Pascal: 566 µs
* scalar ASM: 610 µs

So the SIMD digit reduction still seems to be a very good fit for `GetExtended()`.

For `GetNumericVariantFromJson()`, however, we consistently see the opposite result.

Delphi Win64:

* pure Pascal: 579 µs
* scalar ASM: 730 µs
* SSSE3 ASM: 925 µs

FPC/Linux x64:

* pure Pascal: 590 µs
* scalar ASM: 614 µs
* SSSE3 ASM: 902 µs

On this more JSON-oriented input distribution, the SSSE3 front-end therefore seems to cost more than it saves. The pure Pascal code performs particularly well, probably because its fast path accumulates the first digits directly and common short JSON numbers don't benefit enough from the fixed SIMD setup/reduction cost.

This does not necessarily contradict the benchmark numbers originally reported in your PR. The input distribution is likely very important here, and at the moment we cannot reproduce the original benchmark because we don't have the actual input data used for it.

Would it be possible to share the original benchmark dataset, or at least some details about its number distribution?

In particular, it would be useful to know:

* the proportion of integers, fractions, and scientific notation;
* typical number lengths;
* the proportion of 1–8 digit vs 9–15+ digit mantissas;
* whether the benchmark numbers came from a real JSON payload, and if that sample could be shared.

It would be very useful to rerun the current implementations against exactly the same dataset.

With our present mixed benchmark, the direction seems to be:

* define `GetExtended()` and `GetNumericVariantFromJson()` as global variables to functions, to be selected at runtime (not compile time)
* keep the x64 SSSE3 implementation for `GetExtended()` is SSSE3 is available - fallback to the pascal version which is faster even than the scalar asm from my tests;
* use the pure Pascal version for `GetNumericVariantFromJson()` by default - but for specific process, the x64 ASM implementation could be enabled at runtime for comparison or workloads dominated by longer numbers.

But before making the latter choice final, reproducing the original PR workload would help determine whether this is simply a different crossover point caused by the input distribution.

I remember that Lemire himself did fallback to a scalar algorithm for integer parsing, because from its extensive tests SIMD was not beneficial from realistic JSON input data for such numbers...

synopse:
I have made the runtime redirection via global functions:
https://github.com/synopse/mORMot2/commit/acc10bf52

One amazing outcome of this work is that several ideas from the PR also led us to simplify and speed up the pure Pascal parsers.

So even if the SSSE3 variant parser does not become the default for JSON, the PR still resulted in improvements to the portable code as well, especially in the rewritten Pascal `GetNumericVariantFromJson()`.

Moonbot-Tech:
This task is quite addictive :) It turned out to be much more complicated and interesting than the initial "what could be simpler than translating an existing algorithm into ASM?"

The practical gains from micro-optimization have become secondary, but the new details are still fun to investigate :)

For example, while looking into your results, we ran into a store-to-load forwarding effect. On a Ryzen 7 5800X with Delphi 12.2, the JSON SIMD parser does lose in your benchmark: **14.71 ns per call versus 11.06 ns for Pascal**.

But if we prepare all three states of each string in advance, keeping the same numbers and call order, a common conversion loop gives **5.93 ns for SIMD versus 10.14 ns for Pascal**.

In the original test, individual bytes are changed immediately before the calls: the first digit, the decimal point, and the terminating zero. The subsequent 16-byte SIMD load incurs a penalty. We checked this separately: even writing the same value back to a single byte adds about 8–10 ns to SIMD, while Pascal is affected much less.

Our original corpus contains numbers extracted from 780 saved public REST/WS exchange responses, preserving duplicates and order. Attached are the numbers, their distribution statistics, and a test to reproduce this effect. It would be interesting to see how it behaves on your CPU.

[mormot-pr582-corpus-and-repro.zip](https://github.com/user-attachments/files/32293816/mormot-pr582-corpus-and-repro.zip)

synopse:
I guess that when we reach such low level L1 cache issue, it sounds like if the code is already very fast, whatever code (asm or pascal) we use! :dancers:

synopse:
(edited)

I think the store-forwarding observation is interesting, but I'm not sure it explains our actual workload. More importantly, I tested the complete mORMot JSON regression benchmark with both the Pascal and SSSE3 implementations. The timings are essentially identical; the few-percent differences are within normal CPU/system noise.

So although the SSSE3 implementation is clearly faster in the isolated GetExtended() benchmark, this does not translate into a measurable improvement in actual JSON parsing. GetExtended() and GetNumericVariantFromJson() are called only after the JSON value has already been identified (by type or attribute name), and prepared (e.g. with an ending 0 before GetExtended call). At this point I am concerned that we are optimizing a component which is no longer a bottleneck in the real parser.

I'd therefore prefer to keep both Pascal and SSE3 implementations, and let decide at runtime which one to prefer if it makes any difference, unless we can demonstrate a measurable end-to-end benefit on realistic workloads - not just micro benchmarks like we did here.

Moonbot-Tech:
Let me clarify what I meant! Your test contains this sequence:

```pascal
p[1] := '7';
PCardinal(@value.VType)^ := varEmpty;
GetNumericVariantFromJson(p + 1, value, true);
```

The write to `p[1]` does more than add its own execution cost — **it also slows down the subsequent load inside the SIMD parser**. Even when writing the same byte: the fresh store may still be in the store buffer, while SIMD loads 16 bytes. Running in one thread with the data in L1 does not rule this out.

Preparing the input beforehand removes that immediate dependency:

```pascal
p[1] := '7'; // before timing

Timer.Start;
for i := 1 to N do
begin
  PCardinal(@value.VType)^ := varEmpty;
  GetNumericVariantFromJson(p + 1, value, true);
end;
```

That is why Python prepares the input buffers from your test numbers in advance. **The entire timed loop and the timer are inside Delphi**; Python is not involved in the conversion loop.

During ordinary numeric parsing through `TDocVariant`, the input is not overwritten like this immediately before conversion. In our case with quoted numbers, inserting `#0` is also separated from conversion: we first index the document, then read the fields through `GetExtended`.

With your same numbers in the same order, but prepared beforehand, we measured **5.93 ns for SIMD versus 10.14 ns for Pascal**. So your plan to test an unchanged buffer is exactly what I meant :)

synopse:
I With `TDocVariant`, the input is overwritten like this, but the conversion itself is not the bottleneck. The surrounding work — in particular memory allocation and property/name handling — is much more significant. With RTTI, `GetExtended()` is called just after the in-place `#0` modification, but again this does not seem to have a measurable impact on the complete JSON parsing time.
  From the mORMot perspective, I care much more about **wall-clock time, requests per second, and CPU/process load over a long-running workload** than about a few nanoseconds measured by a microbenchmark on pre-loaded data.
  I have therefore also compared the Pascal and SSSE3 implementations in the actual mORMot JSON regression tests. The complete JSON parsing timings are essentially identical; the few-percent differences are consistent with normal CPU/system noise and power-management effects.
  For example, on the same 1 MB JSON input:
  - `TDocVariant`: **241.7 MB/s**
 - `TDocVariant dvoIntern`: **261.4 MB/s**
 - `TDocVariant FromResults`: **562.4 MB/s** for `exp`, **366.1 MB/s** for `non exp`
 - `DynArrayLoadJson`: **525.3 MB/s** for `exp`, **353.7 MB/s** for `non exp`
 - `TOrmPeopleObjArray`: **449.6 MB/s** for `exp`, **343.9 MB/s** for `non exp`

  These are complete JSON parsing workloads, not isolated number-conversion loops. And replacing the Pascal conversion by the SSSE3 implementation does not produce a corresponding wall-clock improvement in these tests.
  So I am now quite confident that both `GetExtended()` and `GetNumericVariantFromJson()` are already **very fast**, including the Pascal implementation. `GetExtended()` reaching around 1 GB/s in isolation is already a very good result.
  For a specific workload containing very large numbers, where numeric conversion represents a significant fraction of the total processing time, the SIMD implementation may certainly be useful. I don't dispute that.
  My concern is simply that we should not optimize the microbenchmark at the expense of code complexity when the optimization is not measurable in the actual JSON parsing workload. **For mORMot, the wall clock is ultimately what matters.**

 **So I would favor keeping both Pascal and SSSE3 implementations as they are, and let the user select at runtime or compile time what is needed for an actual kind of dataset.**

Moonbot-Tech:
Yes, we agree that the impact on complete JSON parsing is modest. Our mixed corpus benchmark, including building, traversing and freeing `TDocVariant`, did not show a consistent improvement either.

The difference becomes more noticeable with a lighter representation of the result. We use `GetJsonPropName` / `TGetJsonField` to parse directly in a mutable buffer, keep a compact structural index, and then read the required numeric fields through `GetExtended`. There are no separate string allocations or Variant-tree nodes for each field. In our earlier measurements, this complete order-book reading cycle was about 8–9% faster for Binance Spot and Bybit Linear. That comparison was against the Pascal version at the time; we are not assuming those percentages carry over to the latest upstream changes.

For frequent short messages, such as Binance trades, we use an even more direct path: a byte scan of the JSON, calling `GetExtended` on pointers to the price and quantity, without a DOM, structural index or temporary copies of the numeric strings. On numbers with that JSON suffix, the conversion speedup is clearly visible. This is the kind of handler where a fast `GetExtended` is useful to us.

So keeping both implementations with runtime selection makes good sense. Our clarification was mainly about why SIMD lost in the original microbenchmark, rather than a claim of a large speedup for every JSON parser.

synopse:
Small correction and final benchmark update.

I rewrote `BenchmarkNumbers` (see my above commit) so that 200,000 values are generated beforehand into a single ~1.7 MB contiguous `#0`-separated buffer which remains untouched during the timed loops. This removes the immediate store -> SIMD load dependency discussed above.

With the corrected benchmark, SSSE3 clearly wins for both functions.

On FPC/Linux x64, without `Check()` calls inside the timed loop:

* `GetExtendedPas`: about 3.0 ms
* `GetExtendedSsse3`: about 2.0 ms
* `GetNumericVariantPas`: about 2.46 ms
* `GetNumericVariantSsse3`: about 1.85 ms

So SSSE3 is roughly 25% faster in elapsed time for `GetNumericVariantFromJson()` on this mixed corpus, and about one third faster for `GetExtended()`.

Delphi Win64 shows the same overall trend.

This also means the store-forwarding observation was relevant: once the input mutations were removed, the large SIMD regression disappeared. I still don't reproduce quite the original ~2x gain, which is probably explained by CPU/compiler and corpus distribution, but the SIMD advantage is now clear and repeatable.

Given these results, I plan to select the SSSE3 implementations by default on capable x64 CPUs for both `GetExtended()` and `GetNumericVariantFromJson()`, while keeping the optimized Pascal versions as portable fallbacks/reference implementations.

So I would characterize the journey as a successful optimization investigation rather than merely a successful PR merge. The code that finally emerged is better because the initial performance claim was questioned, reproduced imperfectly, contradicted, re-examined, and finally validated under a cleaner benchmark.
Thanks again @Moonbot-Tech for your amazing input and feedback!

Issue #588, opened 2026-09-16
Port LogViewer tool from mORMot 1 to mORMot 2


synopse:
## Port mORMot 1 LogView to mORMot 2

Port the existing mORMot 1 **LogView** application to **mORMot 2**.

A first implementation could target **FPC/Lazarus LCL** as a native desktop application, providing a straightforward path to cross-platform support while keeping the existing LogView workflow and functionality.

### Initial goals

* Port the current mORMot 1 LogView functionality to mORMot 2 APIs and types.
* Prefer a native **LCL** implementation for the first version.
* Keep compatibility with existing mORMot log files, including the current mORMot 2 log format.
* Support mORMot 2-specific information which the old viewer cannot fully expose, such as newer CPU/CPUID information on Intel and ARM.
* Ensure correct handling of remote log entries, including the case discussed in #344.

### Target platforms

Official binaries should eventually cover at least:

* Windows x64
* Linux GTK x64
* Linux GTK aarch64
* macOS aarch64

Win32 is not a priority, especially because memory-mapped log files are constrained by the 32-bit address space.

**Nice To Have:** Delphi VCL port of the app, if it is feasible easily, once the LCL app is stabilized.

This port could also provide the proper long-term resolution for #344, rather than adding compatibility workarounds to the old mORMot 1 viewer.

landrix:
Here is a first version of the LogView port, before opening a PR:
https://github.com/synopse/mORMot2/compare/master...landrix:mORMot2:feature/logview-588

Before going further, I would like your opinion on the only core change, in `mormot.rest.http.server.pas`:
https://github.com/landrix/mORMot2/commit/92c7e314d

```pascal
TRemoteLogReceivedFrom = procedure(const Text: RawUtf8;
  ConnectionID: TRestConnectionID; const RemoteIP: RawUtf8) of object;

constructor TRestHttpRemoteLogServer.CreateWithSender(const aRoot: RawUtf8;
  aPort: integer; const aEvent: TRemoteLogReceivedFrom);
```

**Why:** as you suggested for #344, LogView fixes the rows when they are received: every row is normalized to one layout before `AddInMemoryLine()`. But two clients connected at the same time both send e.g. thread 1, and their rows would be mixed into a single thread. With the connection ID, LogView maps each client's thread numbers to its own synthetic threads, named after the sender (e.g. `127.0.0.1 #3 Worker0`).

[...]

PR #589, opened 2026-09-16
core: fixed TSynLog deadlock when the log file can't be created

landrix:
When the log file can't be created (e.g. the `DestinationPath` folder does not exist), all threads except the first one block forever at their first log call.

Conditions:
- `PerThreadLog = ptIdentifiedInOneFile` (the default)
- `sllInfo` and `sllException` in `Level` (e.g. `LOG_VERBOSE`), so this family is `HandleExceptionFamily`
- the file creation fails

### What happens
1. The first log in a thread calls `LogFileInit`. It sets `logInitDone`, then `CreateLogWriter` raises `EOSException` from `TFileStreamEx.CreateFromHandle`.
2. Before `CreateLogWriter` handles it with `except`, `SynLogException` intercepts it. `tiExceptionIgnore` is not set, so it calls `LockAndDisableExceptions` on the same instance. `LogFileInit` is skipped because `logInitDone` is already set, then `SynLogGlobalLock.Lock` runs, then `AddLogThreadName` calls `LogHeaderNoRecursion(fWriter = nil)`, which raises an AV.
3. The AV happens before the `try..finally` of `SynLogException`, so this `Lock` is never released. The thread keeps working because the critical section is recursive. Every other thread waits on `SynLogGlobalLock`.

### Fix
`LogFileInit` now sets `tiExceptionIgnore` for the current thread and restores the flags in its `finally`, as `PerformRotation` already does.

### Test
`TTestCoreBase.Debugging` has a new `TestLogFileInit`. It overrides `CreateLogWriter` and checks `ExceptionIgnoreCurrentThread`. Without the fix it fails with one assertion, and the suite does not hang.

### Verified
- Standalone repro (below): hangs without the fix and completes with it, on:
  - FPC 3.3.1 aarch64-win64
  - Delphi 13 Win64
  - FPC 3.3.1 aarch64-linux
- It completes without the fix with an existing folder, with `ptMergedInOneFile`, or without `sllInfo`.
- `mormot2tests`:
  - Delphi 13 Win64: all passed.
  - FPC aarch64-win64: only the pre-existing `TDynArray` failures, identical without the patch.
  - `/loop` on `Debugging`: no failure.

<details><summary>Standalone repro</summary>

```pascal
program synloghang;

{$I mormot.defines.inc}
{$apptype console}

uses
  {$I mormot.uses.inc}
  classes,
  sysutils,
  mormot.core.base,
  mormot.core.os,
  mormot.core.text,
  mormot.core.log;

type
  TWorker = class(TThread)
  protected
    procedure Execute; override;
  end;

procedure TWorker.Execute;
var
  i: integer;
begin
  for i := 1 to 30 do
    TSynLog.Add.Log(sllInfo, 'row %', [i]); // blocks forever without the fix
  TSynLog.Add.NotifyThreadEnded;
end;

var
  w: array[0..2] of TWorker;
  i: integer;
begin
  with TSynLog.Family do
  begin
    Level := LOG_VERBOSE;
    PerThreadLog := ptIdentifiedInOneFile;
    DestinationPath := Executable.ProgramFilePath + 'nonexisting' + PathDelim;
  end;
  TSynLog.Add.Log(sllInfo, 'main');
  for i := 0 to high(w) do
    w[i] := TWorker.Create(false);
  for i := 0 to high(w) do
  begin
    w[i].WaitFor;
    w[i].Free;
  end;
  writeln('done');
end.
```
</details>

Found while porting LogView for #588: a remote logging client process hung whenever its log folder did not exist.


synopse (approved):

synopse:
Thanks for sharing!

I am about to commit a huge refactoring of TSynLog locking, to replace the giant lock with several smaller locks. It could be seen as the first part of it. ;)

Issue #344, opened 2025-02-13
LogView.exe can't display the remote message that contains "thread id" field


zen010101:
more info: https://synopse.info/forum/viewtopic.php?id=7174

synopse:
I looked at https://github.com/synopse/mORMot2/compare/master...landrix:mORMot2:fix/issue-344-remote-log-thread#diff-b4cb57d80cac7fb42f14850dc4439b90beaea105bd08fac4f276806e534d0c22R7912
@landrix 

The line ending for remote lines as nil is a fine fix.

But I don't like the fact that you re-parse the layout of all lines. It slows down the parsing/reading/indexing a lot.
IMHO the remote lines should be fixed when received, if they don't match the expected format.
But the real fix is to ensure the remote lines are in the right format!

zen010101:
Thanks @landrix for picking up this old issue, and @synopse for reviewing the proposed fix.

I'm no longer using remote logging, so unfortunately I'm not in a position to test it at the moment.

For what it's worth, I agree with @synopse's suggestion: keeping the log format consistent at the source or normalizing it on receipt seems preferable to detecting the layout of every line.

I appreciate you both taking the time to look into this.

landrix:
The documentation states that the log viewer is located here: 

https://github.com/synopse/mORMot2/blob/68f8a3cbbe015b40d6ae0ef16282598f9b9501b0/docs/mORMot2-SAD-Chapter-25.md?plain=1#L680

But that is not the case. Should it be ported from mORMot1? What would the requirements for a ported version be?

synopse:
This documentation is no reference, it is AI generated and of course not always accurate. ;)

landrix:
Should logview be ported in any way?

synopse:
Yes, I concur LogView should be ported to mORMot 2 especially with FPC / LCL compatibility so that it could run natively on all targets.
We would need at least Win64 + Linux GTK x64 aarch64 + MacOS aarch64 binaries as official release I guess. Win32 may not be mandatory because the file is mapped in memory so there is a 2GB limitation with 32-bit systems.
I currently still use mORMot 1 LogView because it is still perfectly compatible with the mORMot 2 format (which is backward compatible), but some minor details (e.g. CPUID flags on Intel or ARM) are not available in the old viewer.

landrix:
I’ve revised the branch again and would like to split it into two PRs.

[...]

PR #592, opened 2026-09-18
tools: LogView port to mORMot 2 as a FPC/Lazarus LCL app - #588 and #344

landrix:
This ports the mORMot 1 LogView sample to mORMot 2, as discussed in #588. It also fixes the remote logging display of #344 "when received", as you suggested.

### Commits
1. **`tools: port the mORMot 1 LogView sample as a FPC/Lazarus LCL app`**
   - Adds `src/tools/logview` and an entry in `src/tools/README.md`.
   - The viewer logic is the existing `TSynLogFileView`; the form was ported from `LogViewMain`.
   - Changes from mORMot 1:
     - `TMemo` replaces `SynMemoEx`.
     - Sizes follow the font instead of fixed pixels (row heights, columns, left panel).
     - The `.mab` lookup uses your new `TDebugFile.Lookup`, read-only and `.mab` only.
2. **`rest: TRestHttpRemoteLogServer.CreateWithSender`**
   - The core change you already approved in #588, unchanged.
3. **`tools: LogView remote logging with any client layout - fix #344`**
   - Adds the UI-free `TRemoteLogNormalizer` in `mormot.tools.logview.remote.pas`.
   - Every received row is rewritten to one layout (UTC calendar stamp + thread + level) before `AddInMemoryLine()`.
   - The thread numbers of each HTTP connection are mapped to their own synthetic threads, named after the sender, e.g. `127.0.0.1 #3 Worker0`.

### Why normalize in the tool and not in TSynLogFile
`TSynLogFile` detects its layout once, which is right for a file. A live stream mixes clients with or without thread column, and with calendar, zoned or high resolution timestamps. High resolution rows of several clients can't share one `fFreq` and start time. So those rows get the UTC receipt time, and keep their original counter in the text as `[hires <16 hex>]`. `Enter` rows are left unmarked, to keep the profiler names.

### Known limitations
- A reconnecting client is a new sender, and a connection ID may be reused.
- The thread list shows the latest name of a thread only.
- Rows not recognized as `TSynLog` rows are shown as `info`, in a "no thread column" thread of their connection.

### Tested
| Platform | Compiler | Result |
|---|---|---|
| Windows ARM64 | Lazarus trunk, FPC 3.3.1 | file mode, `.mab` lookup, server mode with live clients |
| Linux aarch64, GTK2 | Lazarus 3.0, FPC 3.2.2 | file mode, server mode with live clients |
| macOS aarch64, Cocoa | Lazarus 4.8, FPC 3.2.4rc1, Xcode 26.6 | compiles, but does not link - see below |
| Linux x64 | - | not tested, no machine available |

**macOS:** the LCL Cocoa units fail at link time with the Xcode 26 linker:

`ld: malformed method list atom 'ltmp5' (cocoawsextctrls.o), fixups found beyond the number of method entries`

Plain mORMot code links fine there (`mormot2tests` and a remote logging test client). So this looks like FPC Objective-C metadata against the new Apple linker, not this code. As a result, two Cocoa-specific spots are untested: the memo selection units (UTF-16, as on Win32) and the owner-drawn events check list.

**Live test:** two real `TRestHttpClient.CreateForRemoteLogging` clients send at once, one with calendar and one with `HighResolutionTimestamp`. Each uses `ptIdentifiedInOneFile` with 3 worker threads, `Enter`/`Leave` and warnings. All rows show with their level, and the threads stay separated per connection. Clear List, Save List and reopening the saved file were also checked.

**Normalizer:** it has a console test program that checks its output against `TSynLogFileView`. It covers all layouts, mORMot 1 uppercase hexadecimal stamps, garbage rows and the thread overflow. It is not part of this PR, as it fits no existing test unit; I can add it to `mormot2tests` if you want.

Two core bugs found on the way are already merged: #589 (TSynLog deadlock) and #591 (plain text search).

### Not included
A VCL/Delphi version. The form unit only uses LCL units, so this would be a later step if wanted.


synopse:
Reviewing your code, I have one concern - but not in your code! :)

TRestHttpRemoteLogServer initialize itself as HTTP_DEFAULT_MODE so it is useHttpAsync on Linux or useHttpApiRegisteringURI on Windows. Perhaps useHttpAsync  on all systems could be just fine, or even the good old useHttpSocket which is good enough for a few connections.

AFAIR the connection ID is using a sequence, so should not be reused with useHttpAsync and useHttpSocket - perhaps another benefit of this. But I guess HTTP_REQUEST.ConnectionId: HTTP_CONNECTION_ID from http.sys is also a sequence.

landrix:
Good point, and it explains something in my tests. On Windows without admin rights, the URI registration fails and `TRestHttpServer` silently falls back to `useHttpSocket`. My Windows runs most likely used that fallback all along (they showed small sequence IDs like `#3`). So the actual server depended on the rights of the user.

I added a 4th commit: `TRestHttpRemoteLogServer` now uses `useHttpSocket` on all systems. I chose it over `useHttpAsync` because your fallback code comments it as "less mature on Win", and it is good enough for a few connections. The doc comments are updated. The http.sys port sharing note is gone, and the connection ID is documented as the 31-bit sequence of the socket server, as you remembered.

Tested again with this commit: two simultaneous clients get their own connection IDs on Windows ARM64, and the GUI server mode works on Windows and on Linux aarch64 GTK2 (which ran `useHttpAsync` before).

If you prefer `useHttpAsync`, or an optional `aUse` parameter, just tell me. It is a separate commit, so it is easy to change or drop.

synopse:
The MacOS linking issue is a real concern... which is not about mORMot I guess.

landrix:
You’d probably have to switch to Xcode 15, right? But I’m not doing that.

synopse (approved):

PR #593, opened 2026-09-18
net: fix EsockEINTR typo breaking POSIX compilation of RecvFrom

SDiestelmann:
`TNetSocketWrap.RecvFrom` refers to `EsockEINTR`, which is not declared anywhere in the tree - the neighbouring `SendTo` correctly uses `ESysEINTR`. Any POSIX target therefore fails to compile since 2.4.16947 (`Undeclared identifier`), on both FPC and Delphi.

Found while syncing our fork to 389ef677a.


synopse:
It has been fixed inbetween.
During the day, the trunk could be unstable like this.
Please use the nightly version (around 2AM Europe time we run our regression tests in our LUTI build farm).

PR #594, opened 2026-09-18
core: fix startup hang of Delphi POSIX apps using integer resource names

SDiestelmann:
On Delphi POSIX targets, the RTL `FindResource()` casts its `ResourceName` parameter into a `string`, so an ordinal name built by `MAKEINTRESOURCE(1)` makes it read at address 1.

The resulting access violation happens during `InitUnits`, where the exception machinery is not ready yet: it ends in a stack overflow, and every Delphi POSIX application using `mormot.core.os` hangs at startup. We hit this on Delphi for Android, where `ResourceExists(MAKEINTRESOURCE(1), RT_RCDATA)` is reached through the unit initialization.

`TExecutableResource.Open` and `ResourceExists` now return `false` for such ordinal names on `DELPHIPOSIX`, since that RTL cannot look them up anyway. Named resources are untouched, and FPC is not affected.

Tested with Delphi 13.1 Florence for Android64 and iOSDevice64/iOSSimARM64.


synopse (approved):

synopse:
Thanks for the feedback!

PR #595, opened 2026-09-18
core: fix swapped short/long resourcestring layout on Delphi POSIX

SDiestelmann:
`_OsLoadResString()` reads the resource string layout the wrong way round: `ShortLen = $ffff` is the marker of the *long* layout (as in `system.pas`), but the code takes that branch to read the *short* one - and then asks for 65535 characters.

So every `resourcestring` was returned from the wrong offset on Delphi POSIX, and `I18nResourceStrings`-like code ran into a stack overflow while the exception it raised needed a resource string of its own.

Tested with Delphi 13.1 Florence for Android64, where `Core base` / `I18nResourceStrings` now passes.


synopse (approved):
Typo during conversion, indeed.

PR #596, opened 2026-09-18
core: enhance ICU detection on modern Android

SDiestelmann:
Two fixes for `TIcuLibrary.DoLoad` on Android, both affecting FPC and Delphi alike:

1. **No `libicudata.so` on current Android.** Since the i18n APEX module, the data is no longer shipped as a separate library, so the existing loop over `libicudata.so.NN` finds nothing and ICU is given up on - although `libicuuc.so` and `libicui18n.so` are still published to apps. They are now opened directly as a fallback, without a version suffix.

2. **Inverted version-suffix test.** After the `ICU_VER[]` loop, `if vers <> ''` guarded the *fallback* scan, so a version that had just been found was scanned for again, and a version that had *not* been found was never looked for. It should be `if vers = ''`.

Without this, `TSynAnsiConvert` cannot convert any codepage on a recent Android (`OS error [256]` on e.g. CP1250, shift_jis, big5).

Tested on a Pixel 6a with Android 17, with Delphi 13.1 Florence for Android64: `Core base` / `Charsets` passes again.


synopse (approved):
I have never tested Delphi for Android.

Thanks for testing and giving some feedback!

PR #597, opened 2026-09-18
core: fix unaligned GlobalTime[] breaking its lock on Delphi aarch64

SDiestelmann:
`GlobalTime[]` was declared as a `packed record`, i.e. with an alignment of 1. The Delphi aarch64 linker did place this variable at an odd address on iOS, and the atomic instruction used by `safe.ReadLock` / `WriteLock` (`TRWLightLock`) then raised an access violation.

Symptom in our test runs: every test method died at its first call into `GlobalTime`, while the summary still reported `Total failed: 0`. The faulting address was symbolized with `atos` and the `.dSYM` down to the odd `GlobalTime` address (e.g. `…101f195fe`).

Dropping `packed` does not change the layout - `TRWLightLock` is pointer-sized, and `cardinal` plus `TSynSystemTime` are naturally aligned within the record - it only lets the linker align the variable itself.

Tested with Delphi 13.1 Florence on the iOS simulator (iPhone 18 Pro, iOS 27) and on an iPhone 17 Pro Max.


synopse (approved):
Both first fields are 32-bit so should be aligned just fine.

synopse:
There is a `assert(SizeOf(GlobalTime) = 128);` at the bottom of the unit...

PR #598, opened 2026-09-18
all: make the Delphi Android target compile, link and run

SDiestelmann:
Everything needed to make the **Delphi for Android64** target compile, link and run - part of a larger effort to use the regular mORMot 2 client on Delphi mobile, instead of the mORMot 1 `SynCrossPlatform*` units.

### `mormot.core.os.delphi`

* `FIONBIO` and `SO_PRIORITY` are not published by the Delphi Android RTL: define them (`$5421`, `12`) for that target.
* `statfs` / `sysinfo` / `sched_getaffinity` and the whole `epoll` block were guarded by `OSLINUX`, although bionic exports the same API: they are now under `OSLINUXANDROID`.
* added `GetSystemProperty()` (via `__system_property_get`), as provided by the FPC Android RTL.

### `mormot.core.os.posix.inc`

* `PosixStatFS()` takes an untyped pointer on Delphi, since `PStatfs` is an FPC RTL type.
* the Delphi Android64 RTL names the inode field `__st_ino`.

### `mormot.net.sock.posix.inc` and `mormot.net.sock`

* bionic exports `getifaddrs`/`freeifaddrs` since API 24 only, while an app may target an older `minSdkVersion` and link against those stubs: they are resolved at runtime with `dlsym`, and no interface is returned on older devices.
* **tagged pointers**: on Android arm64 a heap pointer carries a tag in its top byte, which `free()` verifies. `SetRes()` stored the poll events in bits 60..63 and thus destroyed that tag - releasing such a pointer aborted the process. The events now live in bits 48..51, which are always void since user space addresses use at most 48 bits. The epoll path in `mormot.net.sock.posix.inc` follows suit.

### `mormot.defines.inc` and `mormot.core.os`

* `DOUBLETOSHORT_USEGRISU` for `CPUAARCH64DELPHI`: without it, `str()` yields 15 digits and prints `E-0019`.
* `CPU_ARCH_TEXT` was only `'64'` on Delphi aarch64, because `CPUAARCH64` is an FPC-only conditional - use `ABIA64`.

### `test/test.soa.core.pas`

* `EchoRecord` was excluded through `CPUAARCH64`, which Delphi never defines: `ABIA64` covers both compilers.

### Testing

Delphi 13.1 Florence, Android64, on a Pixel 6a with Android 17: the `CoreUnits`, `ORM` and `SOA` suites run, with 13 failed assertions out of 175 million left - all of them artifacts of our test app running the suites in a background thread (`MainThreadID`).


synopse:
Very nice supporting a new target!
:)

Perhaps CPUAARCH64 should be defined on Delphi in mormot.defines.inc for such targets, instead of ABIA64.

src/core/mormot.core.os.pas
```diff
@@ -690,7 +690,7 @@   EOSException = class(ExceptionWithProps);
       'x64'
     {$else} {$ifdef CPUARM}
       'arm' +
-    {$else} {$ifdef CPUAARCH64}
```
synopse, line 693: weird: CPUAARCH64 is defined in mormot.defines.inc for Delphi :
```
{$ifdef CPUARM64}
      {$define CPU64}
      {$define CPUARM3264}
      {$define CPUAARCH64}
      {$define CPUAARCH64DELPHI}
    {$endif CPUARM64}            
```

synopse (changes requested):
Can you try to use the existing CPUAARCH64 ?

PR #599, opened 2026-09-18
variants: implement DispInvoke late-binding on Delphi aarch64

SDiestelmann:
`DispInvokeNamed()` had no va_list support for the Delphi aarch64 compilers, so late-binding with arguments (`doc.Name(0)`, `doc.Add(...)`) read its parameters from a flat buffer that does not exist there - `TDocVariant` and `BSON` then failed with random values (`Out of range Names[1843262128]`).

Two layouts are needed, both modelled on the existing `DISPINVOKE_SYSVAMD64` code:

* **`DISPINVOKE_AAPCS64`** (Android, Linux): `Params` points to an AAPCS64 `va_list` as described in the "Variable argument lists" appendix of the Arm procedure call standard - general purpose values in 8-byte slots below `GR_Top`, floating point values in 16-byte slots below `VR_Top`, then the stack.
* **`DISPINVOKE_APPLEA64`** (iOS, macOS): Apple does not use that `va_list`. As `System.pas` shows, `TVarArgList` is a plain pointer there, because every variadic argument is passed on the stack in 8-byte slots - so integer and floating point values are read the very same way.

To keep `DispInvokeNamed()` readable, the three va_list cases now share one `DISPINVOKE_VALIST` conditional; the flat-buffer code paths are unchanged.

Tested with Delphi 13.1 Florence: Android64 on a Pixel 6a (Android 17) and iOSSimARM64 (iPhone 18 Pro, iOS 27) - `Core process` / `TDocVariant` and the BSON tests pass on both.


synopse (approved):
Great finding!

PR #600, opened 2026-09-18
interfaces: external aarch64 stubs for the Delphi Android and iOS compilers

SDiestelmann:
This fills the two `// TODO: use external .o for Delphi ARM64` placeholders, so `TInterfacedObjectFake` (SOA client side) and `CallMethod()` (`RawExecute`, server side and callbacks) work with the Delphi LLVM aarch64 compilers - for **Android and iOS**.

### The ABI difference

Delphi's aarch64 compiler follows AAPCS64 and passes a by-ref result in the dedicated `X8` register, so it does not consume `X1` as FPC does. `DELPHI_AARCH64_RESULT_X8` covers the three places that follow from it: the stub saves `X8` into `TFakeCallStack.Frame`, no integer register is reserved for the result when the arguments are laid out, and `RawExecute()` puts the result pointer into the new `TCallMethodArgs.ResultX8`.

### The external object

`res/static/delphillvm/delphi-aarch64.s` provides `aarch64fakestub`, `aarch64callmethod` and the VMT entry points, in the style of the existing `delphi-linux-x64.s`. One source, three objects, built by `compile.sh` with the clang of the NDK shipped with the RAD Studio Android SDK: ELF for Android, Mach-O for the iOS device and for the simulator (they differ only in `LC_BUILD_VERSION`). The `#if defined(__MACH__)` block deals with the Mach-O flavour - leading underscore on symbols, no ELF symbol attributes, `L` local labels. The `.cfi_*` directives are as mandatory as in the x64 file, since Delphi uses DWARF exceptions.

### No JIT on iOS

iOS terminates any app that runs code it wrote itself: the JITted VMT stubs live in an anonymous `rwx` page, and the kernel reports `EXC_BAD_ACCESS` / `KERN_PROTECTION_FAILURE` with the `CODESIGNING` termination namespace and "Invalid Page" - the app dies at its first interface call. The simulator does not enforce code signing and hides this.

So under `DELPHI_AARCH64_STATIC_FAKEVMT` the VMT entry points come compiled in: `aarch64fakethunks`, 128 entries of 8 bytes (`mov x16,#i` + `b aarch64fakestub`), and `Compute_FAKEVMT` just points into that table. A `{$message fatal}` guards `MAX_METHOD_COUNT` against the table size. Android uses the same path, and no longer needs executable JIT memory either.

One catch worth knowing: Delphi wraps an `external ... name` procedure into a thunk of its own, so `@aarch64fakethunks` yields that wrapper rather than the table - hence the small `aarch64fakethunkbase` accessor (`adr x0, table; ret`).

### `OSDARWINARM` for Delphi

Apple deviates from AAPCS64 for arguments passed on the stack: they are packed to their natural size (an `integer` takes 4 bytes, not 8). `mormot.core.interfaces` already handles that for FPC through `OSDARWINARM`, which was never set for Delphi - `Client side` failed 72 assertions (`StackIntMultiply(1..10)` returned 0 instead of 3628800) until it was.

### Testing

Delphi 13.1 Florence, with the `SOA` suite (`Interface stubs`, `Mocks and stubs`, `Direct call`, `Client side`, `Server side`, interface callbacks):

* Android64 on a Pixel 6a (Android 17)
* iOSSimARM64 (iPhone 18 Pro, iOS 27) and iOSDevice64 on an iPhone 17 Pro Max

Not covered, and therefore left alone: macOS and Windows on aarch64 - they would need their own object built from the same source.


synopse (approved):
Thanks for giving the .o too because compiling would be difficult for me.

PR #603, opened 2026-09-19
core: fixed TSynLogFile/TSynLogFileView edge cases found by the LogView tool

landrix:
These are points 1-3 of the list I posted in #588. All of them still reproduce on master (2.4.17002): 1 and 2 as an access violation, 3 as a wrong result.

### Fixes
1. `TSynLogFileView.GetLineForClipboard` indexed `fLevels = nil` on a plain text file (no `TSynLog` header). It now returns the raw line, like `LineContains()` does since #591.
2. `TSynLogFile.SetLogProcMerged(true)` on a log with no `sllEnter` at all: the `repeat` loop ran once with `fLogProcNaturalCount = 0`, so it dereferenced `P = nil` and `@fLogProcMerged[0]` of an empty array. It now only merges when there is something to merge.
3. `TSynLogFileView.Select` reset `fSelectedCount` only inside `if integer(fEvents) <> 0`, so unchecking all levels kept the previous rows selected. The reset now happens in both cases.
4. Same method: `Select` with a plain text file and a non-empty `Events` indexed `fLevels = nil` too. Found by a reviewer while checking 3, so it is in the same commit.

### Tests
One added line in `TestPlainText` and a new `TestNoProcNoEvent` in `TTestCoreBase.Debugging`. They also cover an out of range clipboard index, and toggling the profiler and the levels back.

Without the `mormot.core.log.pas` part of this PR, the suite ends with "Some tests FAILED" and exit code 1. With it: FPC 3.3.1 aarch64-win64 and Delphi 13 Win64 pass. The full Delphi suite is green (184,309,777 assertions); the FPC one shows only the pre-existing `TDynArray` failures of this aarch64 compiler.

### Not in this PR
While reviewing, two older risks in the same area came up, both pre-existing and not triggered by the tool: `EventDateTime` does not check the level or the length of a row, and `fThreads` can be shorter than `Count` after unparsed rows were appended live. Tell me if you want me to look at those too.

Points 4-5 of my list in #588 (thread name ordering, and the two remote logging rows written in local time) are behaviour questions, so I'll wait for your opinion. 6-8 I can send as separate PRs.


synopse (approved):

synopse:
6. Keep as this.
7. Don't delete the .mab file.
8. You can fix.

Thanks a lot!

PR #605, opened 2026-09-19
orm: fixed SQLite batch UPDATE statement cache reused across tables 

dkounal:
## Summary

`TRestOrmServerDB` caches the prepared statement of a "uhex" batch UPDATE (`encPutHexID`, the default `TRestBatch.Update()` encoding without `boNoModelEncoding`) in `fBatch^.UpdateSql`.

In `src/orm/mormot.orm.sqlite3.pas`, `ProcessPutHexID()` recreated this statement only when the field bits changed:

```pascal
if not IsEqual(b^.SimpleFields, Fields) then
begin
  b^.SimpleFields := Fields;
  ...
  Props.Fields.ToCsvText(['update ', Props.SqlTableName, ' set '],
    Fields, '=?', [' where RowID=?'], b^.UpdateSql);
end;
```

The target table was not part of the cache key. The only other reset is the `FillZero(b^.SimpleFields)` in `InternalBatchStop`. But `TRestOrmServerDB.InternalBatchStart` returns `false` for any non-INSERT encoding, so `TRestOrmServerBatchSend` never calls `InternalBatchStop` for UPDATE-only batches.

As a result, a "uhex" UPDATE on table B reuses the statement of a previous "uhex" UPDATE on table A whenever both use the same field bits, i.e. when the updated fields have the same indexes in both tables. The rows of B are then silently written into A at the same RowIDs. B is left unchanged, and `BatchSend()` returns 200 with every `Results[]` = 200.

This happens both within a single batch spanning several tables and across two separate `BatchSend()` calls on the same server, since `fBatch` lives on the server as long as the server does.

## Reproduction

```pascal
type
  TOrmPa = class(TOrm)
  private
    fx, fy: integer;
  published
    property x: integer read fx write fx;
    property y: integer read fy write fy;
  end;
  TOrmPb = class(TOrmPa);

// srv := TRestServerDB.CreateWithOwnModel([TOrmPa, TOrmPb], ':memory:');
// srv.Server.CreateMissingTables;  add rows ID=1,2 with y=0 in both tables

procedure UpdateY(cls: TOrmClass; base: integer);
var
  b: TRestBatch;
  r: TOrmPa;
  i: integer;
begin
  b := TRestBatch.Create(srv.Orm, cls, 1000);
  r := cls.Create as TOrmPa;
  try
    for i := 1 to 2 do
    begin
      r.IDValue := i;
      r.y := base + i;
      b.Update(r, 'y'); // "uhex@Pa" / "uhex@Pb"
    end;
    srv.Orm.BatchSend(b); // returns 200
  finally
    r.Free;
    b.Free;
  end;
end;

UpdateY(TOrmPa, 500); // Pa.y = 501,502 - as expected
UpdateY(TOrmPb, 700); // Pa.y = 701,702 and Pb.y = 0,0 !!!
```

## Fix

`TRestOrmServerDBBatch` now has two more fields, `UpdateProps: TOrmProperties`
and `UpdateFields: TFieldBits`. `ProcessPutHexID()` recreates `UpdateSql`
whenever either of them differs:

```pascal
if (b^.UpdateProps <> Props) or
   not IsEqual(b^.UpdateFields, Fields) then
begin
  b^.UpdateProps := Props;
  b^.UpdateFields := Fields;
  ...
```

Why this change is minimal:
- The generated SQL depends only on `Props` (table name and field names) and on
  `Fields`, so the pair is an exact cache key.
- The UPDATE cache no longer shares `SimpleFields` with the INSERT path.
  `SimpleFields` is set by `InternalBatchDirectOne` for INSERT and zeroed by
  `InternalBatchStop`, so the UPDATE cache no longer relies on those
  side effects.
- `InternalBatchStart`/`InternalBatchStop` and the batch dispatch in
  `mormot.orm.server.pas` are unchanged.
- `UpdateProps` starts as `nil` (the record comes from `AllocMem`), so the
  first UPDATE always builds its statement.
- The statement is still reused for consecutive rows of the same table and
  fields. The only extra cost is one pointer comparison per row.

## Also fixed: rejected "uhex" rows now honor `boRollbackOnError`

`InternalBatchDirectOne(encPutHexID)` returns an HTTP status code, not an ID:
- `ProcessPutHexID` returns `HTTP_SUCCESS`, `HTTP_BADREQUEST`, or
  `HTTP_NOTFOUND` when the SQLite step raises.
- `TRestStorageTOrm.InternalBatchDirectOne` returns `HTTP_SUCCESS` or
  `HTTP_NOTFOUND`.

But `TRestOrmServerBatchSend.ExecuteValue` in `src/orm/mormot.orm.server.pas`
treats any positive value as a success (`result := fValueID > 0`).

For example, if a SQLite trigger rejects one row with `RAISE(ABORT, ...)`, the
batch returns 200 with `Results = [200, 404, 200]`. With `boRollbackOnError`,
the batch is not aborted and the other rows stay written. The regular PUT
path (`encPut`) behaves differently: `EngineUpdate() = false` makes
`ExecuteValue` return false, so the batch raises and the auto-transaction is
rolled back. That row's `Results[]` entry stays `HTTP_NOTMODIFIED`.

The fix is a single expression in `ExecuteValue`:

```pascal
result := (fValueID > 0) and
          ((fEncoding <> encPutHexID) or   // returns an ID
           (fValueID = HTTP_SUCCESS));     // returns an HTTP status
```

- `Results[]` content is unchanged: a rejected row is still reported as 404.
- Without `boRollbackOnError`, the only visible change is that the row now
  counts in the `errors=` field of the `EngineBatchSend` log line.
- This also applies to the in-memory `TRestStorageTOrm` "uhex" path. There, a
  failed `UpdateOne()` now counts as a failure, just as its `EngineUpdate()`
  does for `encPut`.

## Tests

New test `TTestSqliteMemory.BatchUpdate` in `test/test.orm.sqlite3.pas`. It
uses two tables with the same layout (`TOrmBatchUpdB = class(TOrmBatchUpdA)`)
and checks:
1. Two separate "uhex" update batches, first on A, then on B: A keeps its
   values and B gets its own.
2. One batch updating A, then B: both tables are correctly updated.
3. A trigger rejects one row of B, without `boRollbackOnError`: the batch
   returns 200, `Results[1] <> 200`, and the two other rows are written.
4. The same rejection with `boRollbackOnError` and
   `AutomaticTransactionPerRow = 1000`: `BatchSend <> 200` and nothing is
   written.

Results:

| | Before the fix | After the fix |
|---|---|---|
| Delphi 13.1 Win32, `-t TTestSqliteMemory` | `BatchUpdate` 9 / 42 FAILED | 42 / 42 passed |
| FPC 3.2.3 x86_64-linux (native), `-t TTestSqliteMemory` | `BatchUpdate` 9 / 42 FAILED | 42 / 42 passed |
| Delphi 13.1 Win32, full `mormot2tests --nontp` | | 0 / 201,128,729 failed |
| FPC 3.2.3 x86_64-linux, full `mormot2tests --nontp` | | 0 / 204,338,514 failed |

Without the `ExecuteValue` change, the rollback assertion (step 4) fails
(2 / 42) as expected.

Note: `Regexp function` fails 3/3 when `TTestSqliteMemory` is run alone with `-t`, with or without this patch, on both compilers. It passes in the full run, so it seems to depend on an earlier test case and is unrelated to this change.

## Compatibility

- **External databases are unaffected.** `TRestStorageExternal`
  (`mormot.orm.sql.pas`) does not override `InternalBatchDirectSupport`, so it
  inherits `TRestOrm.InternalBatchDirectSupport` (`mormot.orm.rest.pas`), which
  returns `dirUnsupported`. `TRestOrmServerBatchSend.ParseValue` then converts
  "uhex" into a regular `encPut` JSON object and calls `EngineUpdate()`.
- Static and virtual tables of a `TRestOrmServerDB` also return
  `dirUnsupported`, because `TRestOrmServerDB.InternalBatchDirectSupport`
  checks `GetStaticTableIndex(RunTableIndex) = nil`. External tables and
  in-memory static tables are therefore never routed to `ProcessPutHexID`.
- Clients using `boNoModelEncoding` send "PUT" objects and never used this
  code path.
- `TRestOrmServerDBBatch` is an internal record, so no public API changes.


synopse (approved):
Another nice catch.

I am glad to see the ORM batch used in all its corner cases, and the proper fix proposed! 👍 
Thanks a lot!

PR #607, opened 2026-09-19
core: fix CPU_ARCH_TEXT on the Delphi aarch64 targets

SDiestelmann:
The Delphi LLVM compilers get `NOASMBLOCK`, and that block undefines `CPUARM3264`, `CPUARM` and `CPUAARCH64`. `CPU_ARCH_TEXT` switches on `CPUAARCH64`, so its architecture prefix was silently dropped: an Android, iOS or Linux aarch64 build compiled with Delphi reported just `64` instead of `aarch64` - e.g. in the `TSynLog` header, in `Executable.ProgramFullSpec` or wherever `CPU_ARCH_TEXT` shows up.

`ABIA64` is defined in the ABI section, which runs before that undef block, and it is not undefined there, so the constant now switches on `ABIA64`. FPC is unaffected: `ABIA64` is defined for aarch64 there as well.

Verified with Delphi 13 for Android64 (device: Pixel 6a).


synopse (approved):
Not very clean, since we would rather have CPUAARCH64 but this is my fault. In the meantime, it sounds fine.

PR #608, opened 2026-09-19
core: fill Executable information for an Android application

SDiestelmann:
An Android app is a `.so` library loaded by `app_process64`, so `ParamStr(0)` is void: `Executable.ProgramFileName`, `ProgramFilePath` and `ProgramName` end up empty or pointing to nowhere usable, and anything that writes "next to the executable" (`TSynLog` files, `.ini` or pid files) has no valid folder to work with.

When that happens, the information is now taken from what Android does provide:
* the `.so` of the process itself - already resolved into `InstanceFileName` by `dladdr()`;
* the package name from `/proc/self/cmdline` (cut at `#0`, and at `:` for the secondary processes of an app);
* the private folder of the application.

Two points worth discussing:

1. `ProgramFilePath` is set to the **app files folder** (what `Context.getFilesDir()` returns on the Java side), not to the folder of the `.so`. The latter is read-only, whereas log files and the like are expected to be writable. If you would rather keep the `.so` folder here, I am happy to change it - then the files folder would need its own variable to be of any use.
2. The path is built as `/data/user/<uid div 100000>/<package>/files/`, with a fallback to `/data/data/<package>/files/`. The `uid div 100000` is the Android user id, which matters for work profiles and multi-user devices, where `/data/data/` is not the right place.

The function bails out and leaves the previous values untouched whenever the content of `/proc/self/cmdline` does not look like a package name, or the folder does not exist.

Tested with Delphi 13 for Android64 on a Pixel 6a: `TSynLog` writes its files as usual, and `Executable.ProgramName` reports the package name. FPC targets are not affected by the `{$ifdef OSANDROID}` block itself, but would use the same code path if they hit the same situation.


synopse (approved):
Nice.

PR #609, opened 2026-09-19
core: name the library for the mach externals on Delphi Darwin

SDiestelmann:
On the Delphi Darwin targets (iOS, macOS), `external 'mach_task_self';` makes the linker look for a *file* named like the symbol, and linking stops with `ld: file not found: mach_task_self`. Delphi wants the library there, exactly as the neighbouring `mach_absolute_time` / `mach_timebase_info` declarations in the same unit already do with `clib`.

FPC is left untouched on purpose, since the bare externals work there. If you prefer, `clib` could be used for both compilers and the `{$ifdef ISDELPHI}` would go away - I just could not test that on FPC/Darwin myself.

Verified with Delphi 13 for iOS (device and simulator).


synopse (changes requested):
A single version with clib for all platforms (like other API definitions just above) sounds easier/better, please.

SDiestelmann:
Done: a single declaration with `clib` for both compilers, the `{$ifdef}` is gone (rebased on current master).

synopse (approved):
Seems cleaner like this indeed.

PR #611, opened 2026-09-19
net: HTTPS on Delphi Android and iOS through System.Net.HttpClient

SDiestelmann:
Neither the Delphi Android nor the Delphi iOS target has OpenSSL or libcurl available, and the socket layer has no TLS of its own there, so every `https://` request fails on those platforms. What the RTL does offer is `System.Net.HttpClient`, which wraps the TLS stack of the OS (OkHttp/BoringSSL on Android, NSURLSession on iOS).

The PR adds (updated after review):

* **`TDelphiNetHttp: THttpRequest`** in `mormot.net.client` - a single class holding the RTL `THTTPClient` and the whole logic, used as `MainHttpClass` and in `TSimpleHttpClient` when the URI is https. No new unit and no intermediate wrapper any more: `System.Net.URLClient`/`System.Net.HttpClient` are listed *first* in the interface uses clause, so that the mORMot identifiers declared afterwards (e.g. `TUri`) keep precedence;
* **`USEDELPHINETHTTP`** in `mormot.defines.inc`, set for the non-FPC Android and iOS targets unless `USELIBCURL` or `NODELPHINETHTTP` is defined. The `{$ifdef OSANDROID} ... {$else}` layout is kept; Delphi iOS gets its own `{$ifdef IOS}` block, which also sets `NOGSSAPIAUTH`.

Everything sits inside `{$ifdef USEDELPHINETHTTP}`, so no other target is affected. `OnUploadProgress`/`OnDownloadProgress` are not implemented.

One quirk worth knowing: under iOS, `NSURLSession` unpacks gzip/deflate by itself but keeps the `Content-Encoding` header, so the class clears it when the body is no longer packed - otherwise mORMot would try to unpack an answer that is already plain.

Tested with Delphi 13 on Android64 and Android32 (Pixel 6a): HTTPS download, rejected/ignored certificate, REST/SOA client calls over https with SynLZ/deflate, plus the core and SOA regression suites. The previous version was also tested on iOS (simulator and device); the reworked one is compiled for iOS without warning.


src/mormot.defines.inc
```diff
@@ -1156,14 +1156,29 @@
   {$undef USE_WINIOCP}        // disable any Windows-specific code
   {$undef USE_THREADWINIOCP}
 
-  {$ifdef OSANDROID}
+  {$if defined(OSANDROID)}
```
synopse: keep $ifdef

src/lib/mormot.lib.nethttp.pas
```diff
@@ -0,0 +1,208 @@
+/// low-level access to the Delphi RTL System.Net.HttpClient
```
synopse: I would not create its own wrapper class for something which is already a RTL wrapper.
No new unit is needed for this. Just create a `TDelphiNetHttp` single class containing all the logic in a single place.

synopse (changes requested):
Please see my comments: no new unit, everything in a single TDelphiNetHttp class seems easier and cleaner.

SDiestelmann:
Done: no new unit any more - `TDelphiNetHttp` holds the `THTTPClient` and all the logic in a single class, and `mormot.defines.inc` keeps its `{$ifdef OSANDROID}` layout. PR description updated accordingly.

synopse (approved):
Thanks for the proposal!

Do we have almost everything we need to proper client-side native generation with Delphi?

synopse:
I tried to add support of your client class to remote DB connection (could be useful) and our REST client.
(see my above two commits)

Feedback is welcome.

SDiestelmann:
Thanks a lot for merging all of them!

About native client-side generation with Delphi: for Delphi clients we don't need generated code any more. Our apps now share the very same interface and record units (`IInvokable` contracts, records with RawUtf8) with the server, and resolve them at runtime via `ServiceDefine` + `Services.Resolve` - the fake-interface stubs (aarch64 for Android/iOS, and now ARMv7) do the rest. That is why we could retire all our former SynCrossPlatform-generated clients: every app (IM<>SYS, BSYS, L<>SYS, SR<>SYS mobile apps) now runs on the regular mORMot 2 client, on Android 32/64-bit and iOS. So code generation remains useful for non-Delphi clients (TypeScript, Swagger...), but not for Delphi itself.

Your two follow-up commits look fine to me - `TRestHttpClientDelphiNet` is exactly what we had locally, and `TSqlDBDelphiNetConnectionProperties` makes sense. Just a small copy/paste leftover in the latter: the constructor comment still says "initialize the properties for remote access via HTTP using libcurl".

PR #613, opened 2026-09-21
all: make the Delphi Android 32-bit (ARMv7) target work

SDiestelmann:
After aarch64 (#598-#600), this makes the last Delphi mobile target, **Android 32-bit (armeabi-v7a)**, compile and run. It is still needed in the field: many industrial Android devices (scanners etc.) run 32-bit images.

`mormot.core.interfaces` only had FPC asm for `ABIA32`, which Delphi cannot compile for ARM. Following the aarch64 approach:

* **`res/static/delphillvm/delphi-arm.s`** (built into `delphi-android-arm.o` by `compile.sh`, NDK clang): `armfakestub`, `armcallmethod` and 128 compiled-in VMT thunks. ARM (not Thumb) code, so that the VMT entries are plain even addresses; EHABI unwind directives (`.fnstart/.save/.setfp`), so that exceptions raised in `FakeCall` can unwind through the stub. The stub builds the existing `ABIA32` `TFakeCallStack` layout, so no Pascal-side layout change was needed.
* The **ABI**, checked by disassembling Delphi-generated code: VFP hard-float, as `CPUARMHF` already assumes - but a by-ref result (string, dynarray, record...) is passed in **R0 and Self in R1**, exactly as Delphi does on Linux x64. So instead of duplicating the fixups, `DELPHI_SYSVX64_RESULT_FIRST` becomes `DELPHI_RESULT_FIRST` and uses `PARAMREG_FIRST`/`PARAMREG_RESULT` (same values as `REGRDI`/`REGRSI` on SysV x64). Likewise `DELPHI_AARCH64_STATIC_FAKEVMT` becomes `DELPHI_STATIC_FAKEVMT`.

Running the regression suites on the device showed three more 32-bit issues:

* **`mormot.core.json`**: ARMv7 raises an alignment fault on `VLDR`/`LDRD` at an address which is not 4-byte aligned - e.g. a `double`, `TDateTime` or `Int64` field after a `boolean` in a packed record - and `unaligned()` is a no-op on Delphi. `_JS_OneProp` and `JsonLoadProp` now process such `rkFloat`/`rkInt64` fields from/to an aligned copy (`DELPHI_ARM32_UNALIGNED`). Other `unaligned()` places (BSON, datetime, db) are not covered; I found no clean way to make `unaligned()` itself work on Delphi as both read and write target (a record type with `Implicit` operators is no valid assignment target any more).
* **`mormot.core.variants`**: `DispInvoke` read the arguments as a flat buffer, but under the LLVM linker `Params` is `@VAList`, and 64-bit values are 8-byte aligned in the ARM va_list (`DISPINVOKE_AAPCS32`).
* **`mormot.defines.inc`**: `EXTENDEDTOSHORT_USESTR` + `DOUBLETOSHORT_USEGRISU` (Delphi ARM32 went into the Win32 branch, where `FloatToText` gives only 15 digits since `extended = double`), and `CPU_ARCH_TEXT` uses `ABIA32`, as #607 did for aarch64.

Everything is behind `DELPHI_ARM32_STUBS` / `DELPHI_ARM32_UNALIGNED` / `DISPINVOKE_AAPCS32` (Delphi `CPUARM32` only), apart from the define renames.

Tested with Delphi 13 on a Pixel 6a running the app as armeabi-v7a: core, crypto and SOA suites (in-process server over HTTP and WebSockets) - interface stubs and SOA pass, remaining failures are those of the 64-bit build (main-thread assumptions of the test app), plus: BSON (unaligned double, see above) and the x509 catalog tests with the internal RSA (`x509-rs256-int`/`ps256-int` generation returns `cvBadParameter` on each 32-bit run, and only sporadically on aarch64 - not investigated yet). Android64 still passes as before; Android32, Android64, iOS simulator and Win64 compile without warning. I could not compile for Delphi Linux64 here, but the renamed constants have the same values there.


synopse (approved):
The "unaligned" issue is a real one... 
Of course, we could try to define some `function unaligned(var d: double): double;` wrapper similar to the one used in FPC RTL itself, but it could be overkill..

